PR #93, opened 2026-04-24
perf + feat: close the aggregate gap vs Neo4j; harden benchmarks

pdlug:
Builds out the TypeGraph perf suite into a real regression gate, adds a head-to-head harness against Neo4j 5.26, and ships three compiler improvements that the harness surfaced. Every query shape on both sides now asserts a cardinality invariant — silent empty-result or wrong-cardinality regressions throw instead of reporting a falsely fast number.

### bench: expanded perf suite and Neo4j 5.26 head-to-head microbenchmark harness

Scoped aggregate, indexed property filter, temporal `asOf`, fulltext/vector/hybrid search shapes, p95 alongside median, history appended to `reports/history.jsonl`. All shapes assert cardinality invariants. `--scale=N` CLI flag (validated `>= 1`). Existing aggregate benchmark dropped its silent `.limit(20)` so the label matches what it measures.

Standalone package under `packages/benchmarks/neo4j-compare/`. Docker Compose + seed + bench runner, shared xorshift32 seed for byte-identical graph shape, matching methodology (2 warmup iterations, 15 samples, median).

### fix: DDL generator emits real column expressions for indexes in Drizzle factories

Indexes threaded through `createSqliteTables({}, { indexes })` / `createPostgresTables({}, { indexes })` were emitting `(unknown, unknown, ...)` instead of their expression keys because the SQL-chunk flattener mishandled Drizzle column references and `StringChunk` values. Regression test pins the fix.

### fix: persist vector embeddings on SQLite backend
`store.nodes.X.create({ ..., embedding: [...] })` on SQLite was silently dropping the embedding because the backend never implemented `upsertEmbedding`/`deleteEmbedding`. Wires both methods up, opt-in via a new `hasVectorEmbeddings` option (`createLocalSqliteBackend` flips it automatically when sqlite-vec loads). End-to-end integration test.

### perf + feat: count aggregate fast path and Postgres CTE improvements

- **NOT MATERIALIZED** on Postgres traversal CTEs — stops the planner from opaquely materializing the twice-referenced CTEs and picking pathological join orderings on larger graphs.
- **LIMIT push-down** in the count aggregate fast path — when the outer `LIMIT` doesn't depend on aggregate results, push it past the `GROUP BY` into the start CTE. Semantics-preserving.
- **countEdges / countDistinctEdges** — new aggregators that count edges directly, skipping the target-node join when target validity isn't part of the question. `count(targetAlias)` behavior unchanged. Mixed aggregates (both in one query) keep the node join but switch to `LEFT JOIN` with node-side filters in `ON` so edge counts stay accurate.

## Measured impact

TypeGraph SQLite vs Neo4j 5.26 w/ honest 1,200-group aggregates.

| Shape | TypeGraph SQLite | Neo4j 5.26 |
|---|---|---|
| Forward traversal | 0.2ms | 1.1ms |
| 3-hop traversal | 0.5ms | 1.1ms |
| 1000-hop recursive | 1.3ms | 2.1ms |
| Aggregate `count(target)` | 12.6ms | 5.4ms |
| Aggregate `countEdges(e)` | **8.1ms** | — |
| Aggregate DISTINCT | 13.5ms | 6.1ms |

On Postgres the `countEdges` aggregate lands at **3.6ms** — faster than Neo4j's 5.4ms on this shape.

### DIFF
--- a/.changeset/count-aggregate-fast-path.md
+++ b/.changeset/count-aggregate-fast-path.md
@@ -1,0 +1,13 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Push `LIMIT` past `GROUP BY` in the count aggregate fast path when it's safe.
+
+When `groupByNode(...).aggregate({ x: count(alias) })` is paired with an optional traversal and a `.limit(n)` that doesn't depend on the aggregate (no `ORDER BY`, or an `ORDER BY` restricted to group keys), the compiler now emits the `LIMIT` inside the start CTE. The `GROUP BY` runs over `n` rows instead of the full start set — `O(limit)` grouping work instead of `O(|start|)`. When `OFFSET` is also set, it rides along with the `LIMIT` into the start CTE and the outer `SELECT` drops its own `LIMIT`/`OFFSET` so neither clause is double-applied.
+
+The fast path also picks `INNER JOIN` over `LEFT JOIN` for the target-node join whenever a `whereNode()` predicate applies to the target alias, so those predicates constrain every aggregate — including `countEdges(...)`. `LEFT JOIN` remains the strategy when only temporal/delete filters apply to the target, so `countEdges` and `count(target)` can coexist in one query with divergent semantics.
+
+No change to query semantics — aggregate counts still reflect the same `count(target)` as before, including the target node's temporal and deletion filters. No change to aggregate queries without a `LIMIT`. No change on SQLite or PostgreSQL query shapes outside the fast path.
+
+Measured impact: scopes down group-by work for "top-N by count"-style aggregate queries. No impact on the blog-post benchmark's full-graph aggregate (which measures the ungrouped 1,200-user case and intentionally runs without a `LIMIT`).
--- a/.changeset/count-edges-aggregator.md
+++ b/.changeset/count-edges-aggregator.md
@@ -1,0 +1,39 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Add `countEdges(edgeAlias)` and `countDistinctEdges(edgeAlias)` — edge-count aggregators that skip the target-node join in the count aggregate fast path.
+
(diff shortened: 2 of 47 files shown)


pdlug:
Addressed both pre-landing review issues in [`c3f732b`](https://github.com/nicia-ai/typegraph/pull/93/commits/c3f732b).

**P1 — `countEdges` ignoring target-alias predicates** (`src/query/compiler/index.ts`):
The fast path now picks the target-node JOIN kind by whether a `whereNode()` predicate targets the traversal's target alias:
- INNER JOIN with node-side filters in `WHERE` when target predicates exist. Every aggregate — including `countEdges` — is constrained by the predicate, and required traversals whose targets all fail the predicate correctly drop the start row.
- LEFT JOIN with temporal/delete filters in `ON` when no target predicates exist. Mixed aggregates (`count(target)` + `countEdges(e)` in one query) retain their divergent semantics: edge counts include edges to expired targets, node counts exclude them.

**P2 — OFFSET double-applied** (same file):
`startCteLimit` carried both limit and offset into the start CTE but the outer `SELECT` kept emitting them too, so `.limit(20).offset(20)` returned an empty page. The outer clause is now suppressed when push-down applies, and the logical plan is rewritten to drop the `limit_offset` node so the emitter invariant check aligns with what gets produced.

Regression tests (both issues catch themselves if reintroduced):
- `tests/query-builder.test.ts`: asserts INNER JOIN with predicate in WHERE for `whereNode` + `countEdges`, LEFT JOIN + filters in ON for mixed-aggregate no-predicate case, and that outer LIMIT/OFFSET is suppressed after push-down.
- `tests/count-edges-semantics.test.ts`: three runtime tests against a real backend — `countEdges` respects target predicates (Bob-not-Eve), required traversal drops groups whose targets all fail the predicate, and `.limit(10).offset(10)` returns a disjoint page from `.limit(10)`.

Updated `.changeset/count-aggregate-fast-path.md` to describe both correctness fixes alongside the existing LIMIT push-down.

Validation: 2,689 SQLite tests pass (+5 new), 309 Postgres tests pass, lint + typecheck + knip clean.

Issue #129, opened 2026-05-09
Refactor: unified TableContribution contract for strategy-owned tables


pdlug:
## Background

#128 fixed the immediate fulltext bootstrap gap with two arms:

- **Arm 2:** `backend.ensureFulltextTable()` narrow-ensure called from `loadActiveSchemaWithBootstrap` (works for both Postgres `tsvectorStrategy` and SQLite FTS5; canonical fix).
- **Arm 1:** Typed Drizzle pg-core `tables.fulltext` for `tsvectorStrategy` so drizzle-kit consumers see the table in their migration files (Postgres only — FTS5 virtual tables aren't Drizzle-modelable).

That fix works but reveals a deeper inconsistency in how typegraph tables are surfaced. The "what tables does TypeGraph need?" answer is now split across **four** categories:

1. **Drizzle named exports** (`nodes`, `edges`, `uniques`, `schemaVersions`, `embeddings`, plus the new `fulltext` for Postgres) — first-class, drizzle-kit picks them up via `export *`.
2. **Drizzle via tables-factory recursion** (`indexMaterializations`, `kindRemovals`, `reconciliationMarkers`) — drizzle-kit recurses into the `tables` object and finds them.
3. **Strategy-owned raw DDL** (SQLite FTS5 fulltext, future pg_trgm / ParadeDB / pgroonga Postgres strategies) — invisible to drizzle-kit; created at runtime by the bootstrap probe.
4. **Per-deployment derived indexes** (the `indexes` option to `createPostgresTables` / `createSqliteTables`) — Drizzle, but consumer-supplied and dynamically generated.

Adding a new strategy-owned table (or a non-Drizzle-modelable storage table for some future feature) without remembering to also wire up an `ensureXTable` method + bootstrap probe re-opens the same trap #128 closed. The convention is documented but not enforced.

## Proposal

A unified `TableContribution` contract that every table (Drizzle or raw-DDL) flows through. Strategies declare their tables via the same shape as the base schema:

```ts
type TableContribution =
  | { kind: \"drizzle-pg\"; table: PgTableWithColumns<any> }
  | { kind: \"drizzle-sqlite\"; table: SQLiteTableWithColumns<any> }
  | { kind: \"raw-ddl\"; tableName: string; createDdl: readonly string[] };

interface FulltextStrategy {
  // existing: matchCondition, rankExpression, snippetExpression, …
  ownedTables(primaryTableName: string): readonly TableContribution[];
}

[...]

PR #144, opened 2026-05-19
feat: createVerifiedStore + assertSchemaCurrent runtime attach; fix #143 contribution-DDL ordering

### LINKED ISSUE #143: 0.26: loadActiveSchemaWithBootstrap runs contribution DDL before the breaking-change check → wrapped query error instead of MigrationError on pending migrations
On a DB with a pending breaking migration, `createStoreWithSchema` → `loadActiveSchemaWithBootstrap` calls `ensureRuntimeContributions` (fulltext contribution DDL) against the **stale persisted** schema **before** `ensureSchema` computes the diff and throws `MigrationError`.

### Where

- `loadActiveSchemaWithBootstrap` (`dist/chunk-4TJI44WP.cjs:2363`) runs `backend.ensureRuntimeContributions(graphId)` unconditionally, then returns the active row.
- `ensureSchema` calls `loadActiveSchemaWithBootstrap` first (`:2407`) and only afterwards computes the schema diff and throws `MigrationError` for breaking changes (`:2458`).
- `ensureRuntimeContributions` → `ensureMarkerTable` + `materializeOne` (fulltext) in `dist/chunk-TVUARC7S.cjs:1160`.

### Symptom

When the persisted active schema is behind the code graph by a breaking change, the fulltext contribution DDL is computed for the new graph but applied against the old table shape. One statement fails, the transaction aborts, and every subsequent statement in the same tx is rejected. The error that propagates is a `DrizzleQueryError` wrapping an idempotent no-op:

```
Failed query: CREATE TABLE IF NOT EXISTS "typegraph_contribution_materializations" ( ... )
```

That table already exists and is healthy (its marker row is present with no `last_error`) — it is collateral damage from the aborted transaction, not the real failure. The originating Postgres error (effectively "current transaction is aborted, commands ignored until end of transaction block") is only reachable via `error.cause`, not the message.

Because the surfaced error is `DrizzleQueryError` and not `MigrationError`, a consumer using the documented migrate-on-`MigrationError` recovery pattern never gets a clean `MigrationError`. The first request after every breaking schema change 500s. It self-heals only because a later/concurrent boot happens to win the migration race (we can see v(N+1..M) schema_versions rows written within the same ~2s as the failed request, with the final version marked active).

### Reproduction

1. Boot a Postgres-backed store and commit graph schema vN.
2. Add a breaking change to the graph (e.g. a new node + edges) so the code graph is vN+1.
3. Cold-boot `createStoreWithSchema` against the vN DB. First boot throws `DrizzleQueryError` (CREATE TABLE on `typegraph_contribution_materializations`) instead of `MigrationError`; a subsequent boot migrates and succeeds.

Environment: `@nicia-ai/typegraph@0.26.0`, Postgres backend (`createPostgresBackend`), `tsvectorStrategy`, Neon serverless driver.

### Asks

1. **Ordering:** run the breaking-change / `MigrationError` check (or at least `getActiveSchema` version comparison) **before** runtime-contribution materialization, so consumers get a clean `MigrationError` and their migrate-then-retry path works on the first request.
2. **Diagnosability:** do not report a no-op idempotent `CREATE TABLE IF NOT EXISTS` as the failure. Surface the originating aborted-transaction cause (e.g. include `error.cause` / the first failing statement in the thrown error message).

### Consumer workaround (for anyone hitting this)

Recover from non-`MigrationError` boot failures by probing `getActiveSchema` + `getSchemaChanges` (both SELECT-only — they do not re-run contribution DDL) and migrating from `active.version` when a diff is pending, then retrying `createStoreWithSchema`.

pdlug:
Two related changes that together unlock running the TypeGraph runtime under a least-privilege, DML-only database role.

Closes #143.

## #143 — `MigrationError` masked by wrapped contribution DDL (patch)

**Symptom.** On a Postgres DB pending a breaking migration, `createStoreWithSchema` 500'd with a `DrizzleQueryError` wrapping a no-op `CREATE TABLE IF NOT EXISTS typegraph_contribution_materializations` instead of a clean `MigrationError`. Consumers using the documented migrate-on-`MigrationError` recovery path never saw it; only a concurrent migration race healed it.

**Cause.** `loadActiveSchemaWithBootstrap` called `ensureRuntimeContributions` (fulltext contribution DDL, derived from the new code graph) **before** `ensureSchema` computed the diff. Against a stale schema the new DDL hit the old table shape, poisoning the surrounding transaction; the error that escaped was the next idempotent statement (the marker-table CREATE), not the originating one.

**Fix.** The loader is now strictly read+bootstrap. `createStoreWithSchema` remains the single canonical durable-marker writer and runs that step **after** `ensureSchemaImpl`, so the breaking-change gate is reached first. Pre-#129 `ensureFulltextTable` fallback is preserved at the canonical writer (factored into `materializeRuntimeContributions`).

**Test.** New regression in `tests/fulltext-bootstrap.test.ts` spies on `ensureRuntimeContributions` across a v1 → v2-breaking cold boot and asserts `MigrationError` is thrown without contribution DDL being attempted.

## `createVerifiedStore` + `assertSchemaCurrent` (minor)

New SELECT-only entrypoints — the runtime counterparts of `createStoreWithSchema` for the least-privilege deployment model:

- **`createVerifiedStore(graph, backend, options?)`** returns `Promise<[Store<G>, SchemaValidationResult]>`. Zero DDL. Throws `MigrationError` on any drift (safe or breaking — the least-privilege runtime cannot migrate), `ConfigurationError` when no schema has been initialized, `StoreNotInitializedError` when the schema is current but runtime-contribution markers are missing/stale.
- **`assertSchemaCurrent(backend, graph)`** — the same gate as a standalone predicate for readiness probes / healthchecks.

Recommended deployment shape is now:

1. **Migration step** (privileged role, DDL/CREATE): `createStoreWithSchema()` once, or `generate*MigrationSQL()` + a one-shot `createStoreWithSchema()`.
2. **Runtime** (least-privilege, DML-only role): attach with `createVerifiedStore()`. Schema drift fails fast with a clean `MigrationError` instead of leaking into hot-path operations or 500'ing on a permission error.

Internal: factored a pure `mergeStoredGraphExtension` helper out of `loadAndMergeGraphExtensionDocument` so the SELECT-only verifier reuses the same parse + extension-merge + deprecated-kind logic without going through the bootstrap-capable loader. `loadAndVerifyGraph` is module-internal; `assertSchemaCurrent` is the sole public verifying primitive in the schema barrel.

**Tests** in `tests/store-verified.test.ts` cover: success returns `unchanged`; fresh DB → `ConfigurationError` without bootstrapping tables (spy-asserted); safe-pending and breaking drift both → `MigrationError` (qualifier in message); missing contribution markers → `StoreNotInitializedError`; zero DDL backend methods on successful verify (spy-asserted); bare `createStore` does **not** catch drift (proves the gate does the work).

## Docs

- `backend-setup.md` — new "Database roles & least privilege" section folds in all three entrypoints; Production example updated.
- `schema-management.md` — "Basic vs Managed vs Verified Store" with a short cross-reference for the role split.
- `troubleshooting.md` — entries for permission-denied-on-boot, `MigrationError` from a verifying attach, `ConfigurationError` on uninitialized DBs, and `StoreNotInitializedError` on the first op.

## Notes for reviewers

- API surface is strictly additive. `createStore` / `createStoreWithSchema` are unchanged.
- `assertRuntimeContributionsInitialized` (already on the backend type) is reused as the zero-DDL marker gate inside `loadAndVerifyGraph`.
- Honest open item out of scope here: `loadAndVerifyGraph` (and `ensureSchema`) compute `getSchemaHash` and, on hash miss, re-`serializeSchema` for the diff. Pre-existing in `ensureSchema`; only affects the drift/failure path. Worth a follow-up to extend the hash cache to also store the `SerializedSchema`, not part of this PR.
- The other unimplemented item from the #143 discussion — unwrapping Postgres SQLSTATE `25P02` ("current transaction is aborted") in `sql-errors.ts` so a poisoned-tx surfaces the originating statement instead of a downstream idempotent no-op — is also out of scope here. The #143 fix already defused the acute recovery-breaking case; the masking is now an operator-ergonomics wart on the privileged migration path, worth a follow-up.

### DIFF
--- a/.changeset/migration-error-before-contribution-ddl.md
+++ b/.changeset/migration-error-before-contribution-ddl.md
@@ -1,0 +1,28 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Surface `MigrationError` before runtime-contribution DDL on a pending
+breaking migration (#143).
+
+`loadActiveSchemaWithBootstrap` ran `ensureRuntimeContributions` (fulltext
+contribution DDL) **before** `ensureSchema` computed the schema diff and
+threw `MigrationError`. Contribution DDL is derived from the current code
+graph, so against a database still on the old schema version it was applied
+to a stale table shape. On Postgres the first failing statement aborts the
+surrounding transaction, and the error that escaped was the idempotent
+marker-table `CREATE TABLE IF NOT EXISTS
+"typegraph_contribution_materializations"` (collateral damage), not a clean
+`MigrationError`. Consumers using the documented migrate-on-`MigrationError`
+recovery pattern never saw a `MigrationError`, so the first request after
+every breaking schema change 500'd until a concurrent boot won the migration
+race.
+
+`loadActiveSchemaWithBootstrap` no longer materializes runtime
+contributions. `createStoreWithSchema` remains the single canonical
+durable-marker writer and runs the materialization step **after**
+`ensureSchema`, so the breaking-change gate is always reached first and a
+pending breaking migration throws `MigrationError` on the first request —
+making the migrate-then-retry recovery path work as documented. The pre-#129
+`ensureFulltextTable` fallback is preserved at the canonical writer. No API
+changes.
--- a/.changeset/verified-store-runtime-attach.md
+++ b/.changeset/verified-store-runtime-attach.md
@@ -1,0 +1,58 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Add `createVerifiedStore` and `assertSchemaCurrent` — the runtime
+counterparts of `createStoreWithSchema` for the least-privilege
+deployment model.
+
+`createStoreWithSchema()` runs DDL (bootstrap, safe auto-migrations,
+durable contribution materialization) and must run under a role with
+`CREATE` privileges. For applications that want their runtime under a
+least-privilege, DML-only role, the previous options were `createStore`
+(zero-DDL attach with no schema gate — drift goes undetected until a
+hot-path operation trips) or hand-rolling a SELECT-only verification
+dance from `getActiveSchema` + `getSchemaChanges`.
+
+This release adds two cleanly named entrypoints that share the same
+zero-DDL verification path:
+
+- **`createVerifiedStore(graph, backend, options?)`** — a SELECT-only
+  attach (zero DDL) with a verification gate. Reads the active schema
+  row and contribution markers, folds the persisted graph extension,
+  and refuses to construct the Store unless the database is at the
+  same schema version as the code graph. Returns
+  `Promise<[Store<G>, SchemaValidationResult]>` mirroring
+  `createStoreWithSchema`. Throws `MigrationError` on any drift (safe
+  or breaking — the least-privilege runtime cannot migrate),
+  `ConfigurationError` when no schema has been initialized, and
+  `StoreNotInitializedError` when the schema is current but
+  runtime-contribution markers (e.g. fulltext) are missing/stale.
+
+- **`assertSchemaCurrent(backend, graph)`** — the same verification gate
+  exposed as a standalone predicate for readiness probes / healthchecks.
+  Returns the `SchemaValidationResult` or throws the same errors.
+
+The recommended deployment shape is now:
+
+1. **Migration step** (privileged role with DDL/`CREATE`): run
+   `createStoreWithSchema()` once at startup, or apply
+   `generatePostgresMigrationSQL` / `generateSqliteMigrationSQL` plus a
+   one-shot `createStoreWithSchema()` to materialize runtime
+   contributions.
+2. **Runtime** (least-privilege, DML-only role): attach with
+   `createVerifiedStore()`. Zero DDL on the runtime path; schema drift
+   fails fast with a clean `MigrationError` instead of leaking into
+   hot-path operations or 500ing on a permission error.
+
+Internal: factored a pure `mergeStoredGraphExtension` helper out of
+`loadAndMergeGraphExtensionDocument` so the SELECT-only verifier reuses
+the same parse + extension-merge + deprecated-kind logic without going
+through the bootstrap-capable loader. No behavior change for the
+existing schema entrypoints.
+
+Documentation: "Database roles & least privilege" in `backend-setup.md`
+now folds in `createVerifiedStore` as the canonical runtime attach;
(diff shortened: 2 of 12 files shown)


pdlug:
Addressed both review findings in `07fe633` (force-push, single commit amended):

**P2 — "generated SQL alone" framing was wrong.** Generated migration SQL only creates tables; it does not write the `schema_versions` row or contribution markers, so the verifying attach would still throw. Reframed across:

- `manager.ts:489-494` — `MigrationError` message in `schemaBehindError` now says: "run `createStoreWithSchema(graph, adminBackend)` under a privileged role (after any generated migration SQL, if you manage DDL externally)".
- `manager.ts:507-510` — `assertSchemaCurrent` JSDoc matches.
- `manager.ts:559-572` — `ConfigurationError` message + suggestion now explicitly note "Generated migration SQL creates the tables but does not initialize the schema row" and direct users to still run `createStoreWithSchema` afterwards.
- `backend-setup.md` "Recommended deployment shape" — the code block previously presented generated SQL and `createStoreWithSchema` as an either/or. Rewritten: `createStoreWithSchema` is mandatory; generated SQL is shown as an optional prerequisite (drizzle-kit users), still followed by `createStoreWithSchema`.
- `troubleshooting.md` — `MigrationError`, `ConfigurationError`, and `StoreNotInitializedError` entries all reframed; `ConfigurationError` entry now explicitly calls out that migration-SQL-alone is insufficient.

**P3 — "zero-I/O" was wrong for `createVerifiedStore`.** Reads SELECTs, so it's zero-DDL / SELECT-only, not zero-I/O. Fixed in:

- `schema-management.md:143-153` — the privileges note now distinguishes `createStore` (synchronous zero-I/O attach) from `createVerifiedStore` (SELECT-only attach, zero DDL — reads active schema row and contribution markers).
- `.changeset/verified-store-runtime-attach.md:20-21` — changeset description matches.
- `backend-setup.md:641` — drift fix: a prettier-vs-markdownlint clash on `_attach_` vs `*attach*` inside the bold lead; dropped the italic to make both happy.

`createStore`'s existing "zero-I/O" claims in `fulltext-search.md`, `errors/index.ts`, and the troubleshooting entry are correct (it really does no I/O) — left as-is.

Verification: typecheck clean, eslint clean, markdownlint clean (3 files, 0 errors), full SQLite suite **3115 passed**, all 9 verified-store tests still green.

Issue #135, opened 2026-05-16
Fulltext storage init: durable, enforced materialization instead of an in-memory per-backend latch on the hot path


pdlug:
## Problem

Strategy-owned **fulltext** table/index DDL is materialized lazily, guarded by an **in-memory, per-backend-instance boolean latch**, and is interleaved into the read/write data path. This is functionally correct today only because the DDL is idempotent and a process happens to keep one backend instance warm — but it is the wrong durability scope, it is inconsistent with how vector indexes are handled, and it **blocks cross-store transaction adoption** (#134).

### What the code shows (0.25.1)

- `backend/drizzle/postgres.ts` (and the SQLite sibling) keeps a per-backend-instance `let fulltextEnsured = false;` latch.  `ensureFulltextDdl()` early-returns when the latch is set; otherwise it runs the strategy DDL and sets it.
- That ensure is wired into **every** fulltext-touching method: `upsertFulltext`, `deleteFulltext`, `upsertFulltextBatch`,  `deleteFulltextBatch`, `fulltextSearch`, and `hardDeleteNode` (the cascade deletes from the fulltext table unconditionally). The in-source comment is explicit that this scatter exists because *"fulltext writes can fire from any code path … so the bootstrap-load probe alone doesn't cover all surfaces."*
- There **is** a proper init point: `loadActiveSchemaWithBootstrap` (`schema/manager.ts`) calls `backend.ensureFulltextTable?.()` on the success path and `bootstrapTables()` on the cold path. But its result is recorded only in the in-memory latch — it is **not durable and not queryable**, so any freshly-built backend instance starts cold and the hot-path guards must defensively re-check.
- **Vector indexes already do this correctly**: they flow through the explicit `materializeIndexes()` verb and the durable `index_materializations` table (`ensureIndexMaterializationsTable` / `getIndexMaterialization`), entirely off the read/write hot path. Fulltext is the only index type with a lazy hot-path safety net *in addition to* its explicit paths.

### Root cause

`createStore` is **synchronous**, so async storage DDL cannot run at construction. Rather than make initialization an explicit, enforced async step, the DDL was deferred onto the first async fulltext operation behind an in-memory latch. The scattered `ensureFulltextDdl` calls are a band-aid for the absence of a durable, enforced "this graph's fulltext storage is materialized" fact.

## Why this blocks #134

[...]

PR #169, opened 2026-06-05
fix(vector): provision per-field embedding storage via durable markers (least-privilege safe)

pdlug:
## Problem

On a least-privilege Postgres role (USAGE on `public`, full DML, but no `CREATE`), **every runtime vector op failed** with `permission denied for schema public` (SQLSTATE 42501). The cause: each vector op (`upsertEmbedding` / `deleteEmbedding` / `vectorSearch` / `createVectorIndex`) lazily ran `CREATE TABLE IF NOT EXISTS` for its per-`(kind,field)` table on the request connection — and Postgres runs the schema aclcheck *before* the `IF NOT EXISTS` short-circuit, so it failed even when the table already existed.

Fulltext already avoided this via the #135 durable-contribution machinery; the #161 vector cutover (v0.29) shipped an in-process-latch shortcut that bypassed it — exactly what the design doc said *not* to do.

## Fix

Vectors now ride the **same** durable-marker machinery as fulltext:

- **Boot (privileged):** `createStoreWithSchema` provisions every embedding `(kind,field)` table + a durable marker; `evolve()` provisions fields it introduces.
- **Runtime (DML-only):** the hot path asserts the marker with a cached SELECT — **never DDL**. An un-provisioned slot throws `StoreNotInitializedError` (not a raw 42501). `createVerifiedStore` verifies vector markers at attach.
- `reembedVectorField` re-stamps the marker after recreating storage at a new dimension; vector-field reclaim clears it when it drops a table.
- The in-process `vectorSlotLatch` is removed; a shared `resolveGraphVectorSlots` enumerator is the single source of truth for boot materialization and runtime assertion.

### Key files

- `backend/drizzle/contribution-materializations.ts` — generalized the materializer to any contribution set; added `ensureVectorSlot` / `assertVectorSlot` / `dropVectorSlot` (+ `force`, `deleteMarker`).
- `backend/drizzle/{postgres,sqlite}.ts` — marker-gated vector methods; shared materializer across tx-scoped backends.
- `store/store.ts`, `schema/manager.ts`, `store/materialize-removals.ts`, `backend/migrate-vectors.ts`, `core/embedding.ts`, `errors/index.ts`.

## ⚠️ Breaking change (minor)

Vector ops now require a prior privileged `createStoreWithSchema` — exactly as fulltext already does. A plain `createStore` + embedding write with no provisioning step throws `StoreNotInitializedError` instead of lazily creating the table.

**Migration:** run `createStoreWithSchema(graph, adminBackend)` once under the schema-owner role after upgrading. It creates the per-field tables + markers; least-privilege runtimes then assert markers (SELECT) and run vector DML with **zero DDL — no `GRANT CREATE` needed.** Changeset included.

## Tests

- New `postgres-vector-least-privilege.test.ts` — the canonical regression: a USAGE-only role (no `CREATE`) runs vector upsert+search after the owner provisions, and gets `StoreNotInitializedError` (not 42501) when unprovisioned.
- Extended the cross-backend materializer unit tests; updated lazy-usage tests/examples/fixtures to provision via `createStoreWithSchema`.
- **SQLite 3493 passed · Postgres + integration 914 passed · property 320 passed** · typecheck + lint clean.

Docs updated (semantic-search, backend-setup, troubleshooting, architecture, graph-extensions, schema-evolution, schema-management) — the old "created lazily on first write" claims were the trap that led here.

### DIFF
--- a/.changeset/marker-gated-vector-storage.md
+++ b/.changeset/marker-gated-vector-storage.md
@@ -1,0 +1,37 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Vector storage now rides the #135 durable-contribution machinery, so the
+runtime never issues DDL on the embedding hot path.
+
+Previously every vector op (`upsertEmbedding` / `deleteEmbedding` /
+`vectorSearch` / `createVectorIndex`) lazily ran `CREATE TABLE IF NOT EXISTS`
+for its per-`(kind, field)` table on whatever connection it executed on. On a
+least-privilege Postgres role (USAGE on `public`, full DML, but no `CREATE`)
+this failed with `permission denied for schema public` (SQLSTATE 42501) — even
+when the table already existed, because Postgres runs the schema aclcheck before
+the `IF NOT EXISTS` short-circuit. The fulltext path already avoided this via
+durable markers; vectors now do too.
+
+What changed:
+
+- **Boot (privileged):** `createStoreWithSchema` provisions every embedding
+  `(kind, field)` table + a durable contribution marker, enumerated from the
+  graph. `evolve()` provisions any embedding fields it introduces.
+- **Runtime (DML-only):** the hot path asserts the durable marker with a cached
+  SELECT and runs DML — never DDL. `createVerifiedStore` verifies vector markers
+  at attach, alongside fulltext.
+- `reembedVectorField` re-stamps the marker after recreating storage at a new
+  dimension; vector-field reclaim (`materializeRemovals`) clears the marker when
+  it drops a table.
+
+**Breaking:** vector ops now require a prior privileged `createStoreWithSchema`
+(exactly as fulltext already does). A plain `createStore` + embedding write with
+no provisioning step throws `StoreNotInitializedError` instead of lazily
+creating the table.
+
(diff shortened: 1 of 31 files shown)


pdlug:
Closing as superseded by #267, which now tracks the underlying bug.

The problem and the design here remain exactly right — vectors must ride the same #135 durable-marker machinery as fulltext, and this PR proved the approach end to end. But the branch has rotted past salvage-by-rebase: 79 commits of drift since its merge-base, with 10 of 31 files in conflict, including every core file (`postgres.ts`, `sqlite.ts`, `contribution-materializations.ts`, `store.ts`, `backend/types.ts`, `errors/index.ts`). Main has since gained the write-pipeline refactor (#188), StoreView (#185), and the transaction-surface raw-SQL/DDL guards (#260), which a reimplementation must integrate with directly rather than merge around.

The branch `fix/vector-least-privilege-marker-gate` is retained as the reference design and test spec — in particular `postgres-vector-least-privilege.test.ts` (the canonical USAGE-only-role regression) ports nearly verbatim. Reimplementation against current main proceeds under #267.

PR #202, opened 2026-07-03
fix: compute facade search top-k over live nodes in SQL

pdlug:
## Bug

`store.search.{vector,fulltext,hybrid}` computed top-k over the embedding/fulltext side tables alone; hydration then dropped tombstoned ids (`fetchNodesByIds`). Under index drift — a side-table row whose node was tombstoned outside the store pipeline (direct `backend.deleteNode` callers, external writers) — drift rows crowd the top-k and results silently shrink below `limit` even when enough live matches exist. The red test drives this to an extreme: 3 drift rows outrank all live rows on both legs, and every search leg returned **0** hits before the fix.

## Fix — one seam, per-engine treatment

A `candidates` node-id subquery (live nodes of the searched kind) is threaded into `VectorStrategy.buildSearch(slot, params, candidates?)` and `buildFulltextSearch(…, candidates?)`. The backend builds it (it owns table names); a follow-up will extend the same argument with compiled `where` predicates.

| Path | Mechanism | Semantics |
| --- | --- | --- |
| pgvector HNSW | `node_id IN (candidates)` + `hnsw.iterative_scan = strict_order` (pgvector ≥ 0.8, probed) | exact on ≥0.8; `ef_search`-bounded below |
| pgvector IVFFlat | `node_id IN (candidates)` + `ivfflat.iterative_scan = relaxed_order` (no strict mode exists) + a `MATERIALIZED` re-sort of the bounded set in the strategy SQL, so relaxed ordering cannot leak misordered rows through the LIMIT | exact on ≥0.8; probe-bounded below |
| sqlite-vec | primary-key `IN (subquery)` pushdown inside the vec0 KNN `MATCH` (spike-verified on v0.1.9) | exact |
| libSQL DiskANN | `vector_top_k` cannot pre-filter → over-fetch 4× + post-filter | recall bounded by headroom (documented) |
| tsvector / FTS5 | plain `WHERE` | exact |

Notes:
- EXPLAIN-verified on pgvector 0.8.1: the filtered form keeps the HNSW index scan (Nested Loop probe of the nodes pkey under the scan).
- The pgvector version probe (`pgvectorIterativeScanSupported`) uses `pg_extension.extversion`, not the GUC — extension GUCs only register after the library loads into a session, so a fresh pooled connection reports the GUC as absent even on 0.8+. Version ≥ 0.8 covers both `hnsw.iterative_scan` and `ivfflat.iterative_scan` (added in the same release).
- The existing `efSearch` SET LOCAL machinery is generalized to a list of parameterized `set_config` overrides with the same snapshot/restore semantics (restore inside adopted transactions, `BEGIN…COMMIT` bundling otherwise).
- Custom `VectorStrategy` implementations that ignore the new optional argument keep the old post-hoc-drop behavior (documented on the interface).
- Hydration keeps its `deleted_at` skip as defense-in-depth for the race window between the search statement and the hydration read.

## Spike correction to the audit

The inline `.similarTo()` CTE shape (join + inner ORDER BY/LIMIT) **does** get the HNSW index scan on pgvector 0.8 — the audit's "brute-force on every engine" claim holds only for the SQLite ANN forms and multi-kind PG unions. Recorded in the design note — the eventual opt-in ANN work for the inline path shrinks accordingly.

## Tests

- New `tests/search-liveness.test.ts` (red-first): drift rows constructed to outrank every live row on both legs; asserts all legs return `limit` live hits, pre- and post-`materializeIndexes`, across local sqlite-vec / libSQL / Postgres.
- `postgres-vector-ef-search.test.ts`: iterative-scan applied on HNSW slots (asserted via drizzle logger bind params) and on IVFFlat slots (`relaxed_order` + the `MATERIALIZED` re-sort wrapper asserted in the statement log), skipped on brute-force slots, no cross-connection leak, restored inside caller transactions, skipped gracefully on pgvector < 0.8.
- IVFFlat drift regression in `search-liveness.test.ts`: a filtered search on a **materialized IVFFlat slot** returns full `limit` live hits in exact similarity order.
- Backend-level vector tests (sqlite backend harness, libsql strategy e2e, legacy embedding migration on both engines) now seed live node rows — required by the tightened `backend.vectorSearch` contract.
- Docs: `semantic-search.md` gains "Deleted nodes never rank" + the DiskANN over-fetch bound.


pdlug:
P1 addressed in 44b5bd94 — implemented the IVFFlat path rather than narrowing the claim.

- `ivfflat.iterative_scan = relaxed_order` is applied transaction-locally on IVFFlat slots (same generalized `set_config` machinery; the pgvector version probe is shared and renamed `pgvectorIterativeScanSupported`).
- IVFFlat has no `strict_order` mode, so the strategy's IVFFlat search SQL re-sorts the bounded candidate set inside a `MATERIALIZED` wrapper (the pgvector-documented pattern) — relaxed ordering cannot leak misordered rows through the LIMIT.
- Regressions: the statement-log test asserts the GUC lands with `relaxed_order` and the wrapper is present; a new drift scenario on a **materialized IVFFlat slot** asserts a filtered search returns full `limit` live hits in exact similarity order.
- Changeset and `vectorSearch` JSDoc updated to state both index-type mechanisms explicitly.

Below pgvector 0.8 both index types remain probe-bounded (documented) — the probe is `pg_extension.extversion`, which also covers `ivfflat.iterative_scan` (added in the same 0.8 release).

PR #222, opened 2026-07-05
Add transaction receipts

pdlug:
Adds `store.transactionWithReceipt()`, an opt-in surface for transaction receipts.

- Adds `TransactionOutcome<T>` and `TransactionReceipt` public types.
- Counts completed write intents at the collection surface for node and edge writes, including bulk input lengths.
- Returns the exact recorded commit anchor allocated by the transaction when `history: true` captured writes.
- Leaves `store.transaction()` untouched; both methods delegate to a shared runner, and receipt proxy wrapping only happens on the receipt path.
- Documents receipt semantics in the store docs and adds a minor changeset.

## Why

Library consumers that project external changes into TypeGraph need a first-class way to know whether a transaction completed TypeGraph writes. Integrators using recorded-time capture also need the exact recorded anchor allocated by that transaction, rather than sampling a graph-wide high-water mark after commit.

Without this API, consumers have to build local wrappers around the transaction collection surface and maintain their own copy of TypeGraph's write-method list. This PR moves that integration seam into the library while keeping it opt-in and neutral: it reports completed write intents and the transaction's recorded anchor, but does not introduce any delivery, subscription, or streaming surface.

## Why a dedicated method (not `transaction(fn, { receipt: true })`)

An earlier revision used an options flag. A flag that changes a method's return type is a pattern this codebase already rejected once (`bulkInsert` exists instead of `bulkCreate({ returnResults: false })`), and it opened a type-soundness hole: receipt options forwarded through a plain `TransactionOptions`-typed variable statically selected the `Promise<T>` overload while the call returned a `TransactionOutcome<T>` at runtime. A dedicated method makes static dispatch and runtime behavior coincide by construction and deleted the runtime option-sniffing machinery outright.

### DIFF
--- a/.changeset/transaction-receipts.md
+++ b/.changeset/transaction-receipts.md
@@ -2,6 +2,7 @@
 "@nicia-ai/typegraph": minor
 ---
 
-Add opt-in transaction receipts that summarize completed collection write
-intents and, for history-enabled stores, return the recorded commit instant
-allocated by the transaction.
+Add `store.transactionWithReceipt()`, which runs a transaction and returns a
+receipt summarizing completed collection write intents and, for
+history-enabled stores, the recorded commit instant allocated by the
+transaction.
--- a/.changeset/transaction-receipts.md
+++ b/.changeset/transaction-receipts.md
@@ -1,0 +1,7 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Add opt-in transaction receipts that summarize completed collection write
+intents and, for history-enabled stores, return the recorded commit instant
+allocated by the transaction.
--- a/apps/docs/src/content/docs/schemas-stores.md
+++ b/apps/docs/src/content/docs/schemas-stores.md
@@ -1323,6 +1323,52 @@ const personId = await store.transaction(async (tx) => {
 // personId is available here
 ```
 
+#### Transaction receipts
+
+Pass `{ receipt: true }` when a caller needs a write summary without wrapping the
+transaction context itself:
+
+```typescript
+const outcome = await store.transaction(
+  async (tx) => {
+    const alice = await tx.nodes.Person.create({ name: "Alice" });
+    const bob = await tx.nodes.Person.create({ name: "Bob" });
+    await tx.edges.knows.getOrCreateByEndpoints(alice, bob, {
+      since: "2026",
+    });
+    return alice.id;
+  },
+  { receipt: true },
+);
+
+outcome.result; // Alice's id
+outcome.receipt.writes; // { nodes: { Person: 2 }, edges: { knows: 1 }, total: 3 }
+outcome.receipt.recorded; // RecordedInstant | undefined
+```
+
+Receipt counts are completed write intents at the collection surface, not rows
+affected:
+
+- Every successful completion of a write method on `tx.nodes.*` / `tx.edges.*`
+  counts. The authoritative method list is `NodeWrites` / `EdgeWrites`.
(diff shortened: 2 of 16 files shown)


pdlug:
## Review + hardening pass (4e988ca)

Full structured review plus an independent adversarial pass. Fixes applied:

**Correctness**
- **Prototype-colliding kind names** (P2): receipt buckets were plain `{}`, and the kind-name rule `/^[A-Za-z_][A-Za-z0-9_]*$/` admits `constructor`, `toString`, `__proto__`. `constructor` produced a string-concatenated count; `__proto__` was silently dropped. Buckets (and snapshots) are now null-prototype, with recorder-level and end-to-end regression tests (a graph with a node kind literally named `constructor`).
- **Bulk-count drift trap** (P3): the per-method count strategy was two hand-maintained `Set`s + one special case; the compile-time `Equal` assertions only pin *names*, so a future bulk method would silently count 1 per call. Replaced with exhaustive `satisfies Record<MethodName, WriteIntentCounter>` maps — a new write method now fails to compile until its count is chosen.
- **Count-after-await** (P3): bulk counts read `args[0].length` after the write resolved, so a caller mutating the input array mid-flight skewed the count. Counts are now pinned at call time (still recorded only on resolve, so rejects count 0).
- **Options allowlisting** (P3): `stripReceiptOption` / `stripRecordedFlushObserver` rebuilt options as `{ isolationLevel }` on *every* transaction; any future `TransactionOptions` field would silently vanish. Both now strip by omission. Added an isolationLevel-passthrough test.
- **Truthy non-boolean `receipt`** (P3): untyped callers passing `receipt: 1` silently got the bare result; now a `ConfigurationError`.

**Documented (design accepted as-is)**
- Cascade/disconnect edge removals happen through the backend, not the edge-collection surface, and do not appear in `edges` (pinned by a new test).
- A rejected bulk write counts 0 even if the backend applied part of it — on SQLite a caught rejection + commit can persist rows the receipt never counted.
- On non-transactional backends, a rejecting callback returns no receipt even though earlier operations committed individually.
- **Known type-soundness limit**: `ReceiptTransactionOptions` widens to `TransactionOptions`, so forwarding receipt options through a plain-typed variable statically selects the `Promise<T>` overload while runtime returns a `TransactionOutcome<T>`. Not fully closable without a separate method (`transactionWithReceipt()`); documented in the JSDoc + docs. Flagging for an explicit accept/split decision.

New coverage includes a full write-surface count test (every node/edge write method once, including `bulkGetOrCreateByConstraint`'s second-argument input — previously the untested lone special case) and `tx.getNodeCollection` counting.

Verified: `pnpm fix`, `pnpm typecheck`, `pnpm test` (4565), `pnpm test:postgres` (1852) — all green.

pdlug:
**Follow-up (224dde5):** resolved the open design question from the review — the `{ receipt: true }` flag is gone in favor of a dedicated `store.transactionWithReceipt()`.

Rationale: a flag that changes a method's return type is the pattern `bulkInsert` already exists to avoid, and the overload-widening hole it created (forwarded options returning a `TransactionOutcome` where the compiler promised `Promise<T>`) is not closable while dispatch depends on a runtime property. The split deletes the mechanism rather than documenting it: `requestsReceipt`, the non-boolean-receipt guard, and `stripReceiptOption` are all gone, `transaction()` is back to two overloads, and both methods share one `#runTransaction`. Net −64 lines.

PR body updated to match. Re-verified: `pnpm fix`, `pnpm typecheck`, SQLite 4564 passed, Postgres 1852 passed.

PR #231, opened 2026-07-07
feat(query): add orderBy() to aggregate queries for top-N by count/sum

### LINKED ISSUE #228: Aggregate queries support limit() but not orderBy() — top-N by aggregate requires JS sorting
## Problem

`ExecutableAggregateQuery` exposes `limit()` but no `orderBy()` (see `packages/typegraph/src/query/builder/executable-aggregate-query.ts`). That makes the most common aggregate question — "top N groups by count/sum" — impossible to express: `limit(n)` without an ordering returns arbitrary rows, so callers must fetch all groups and sort in JS.

## How this was found

`examples/13-aggregates.ts` had a section titled "Top 2 authors by book count" that was actually `.aggregate(...).limit(2)` with no ordering — its output silently cut the actual top author. The example now fetches all groups and sorts in JS, with a comment documenting the limitation.

## Suggested fix

Support `orderBy` on aggregate queries, accepting either a grouped field or an aggregate alias (`ORDER BY COUNT(*) DESC` / `ORDER BY bookCount DESC`), compiled through the shared query path per the dialect-parity rules. `limit` without an explicit order could stay allowed but is only meaningful once ordering exists.

## Notes

- Both SQLite and PostgreSQL support `ORDER BY` on aggregate expressions and on select aliases, so no dialect seam should be needed beyond existing identifier quoting.

pdlug:
Fixes #228. `ExecutableAggregateQuery` exposed `limit()` but no `orderBy()`, so the most common aggregate question — "top N groups by count/sum" — couldn't be expressed: `.aggregate({...}).limit(n)` returned an arbitrary `n` groups, forcing callers to fetch every group and sort in JS (as `examples/13-aggregates.ts` was doing, with a misleading "Top 2 authors by book count" comment on a query that didn't actually order anything).

- Added `ExecutableAggregateQuery.orderBy(key, direction?)`. `key` is fully typed against the fields object passed to `.aggregate({...})` — it can reference either a grouped field or an aggregate alias, and can be chained for multi-key sorts.
- Ordering resolves by referencing the projected SELECT-list output alias (e.g. `ORDER BY "bookCount" DESC`) rather than recompiling the underlying `FieldRef`/`AggregateExpr`. Every `.aggregate()` field is always projected with an alias, so this works uniformly for grouped fields and aggregates, through both the general CTE query path and the compiler's `compileCountAggregateFastPath` optimization (verified with a dedicated integration test), with no dialect-specific handling.
- New `AggregateOrderSpec` AST type kept separate from the existing row-query `OrderSpec` — `OrderSpec.field` is a `FieldRef` threaded through pagination/cursor machinery that aggregate queries never touch, so widening it would have forced dead-branch handling through code aggregate queries can't reach.
- Null ordering for the new alias-based path uses `NULLS FIRST`/`NULLS LAST` syntax rather than the existing `(field IS NULL) direction, field direction` trick used for row queries — verified empirically that referencing an output alias inside a compound expression throws "column does not exist" on PostgreSQL (the alias only resolves when the ORDER BY term is the bare identifier itself).
- Fixed the `examples/13-aggregates.ts` "Top 2 authors" section to actually use `.orderBy("bookCount", "desc")`.

### DIFF
--- a/.changeset/aggregate-order-by.md
+++ b/.changeset/aggregate-order-by.md
@@ -1,0 +1,31 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Aggregate queries now support `.orderBy()`. Previously `ExecutableAggregateQuery`
+exposed `limit()` but no way to order results, so `.aggregate({...}).limit(n)`
+returned an arbitrary `n` groups rather than the top `n` — the most common
+aggregate shape ("top N groups by count/sum") required fetching every group
+and sorting in JS.
+
+`.orderBy(key, direction?)` takes any output name from `.aggregate({...})` —
+either a grouped field or an aggregate alias — and can be chained for
+multi-key sorts:
+
+```typescript
+store
+  .query()
+  .from("Author", "a")
+  .traverse("wrote", "e")
+  .to("Book", "b")
+  .groupByNode("a")
+  .aggregate({ author: field("a", "name"), bookCount: count("b") })
+  .orderBy("bookCount", "desc")
+  .limit(2)
+  .execute();
+```
+
+Ordering resolves against the projected SELECT-list output alias rather than
+recompiling the underlying expression, so it works uniformly for grouped
+fields and aggregates on both SQLite and PostgreSQL with no dialect-specific
+handling.
--- a/apps/docs/src/content/docs/queries/aggregate.md
+++ b/apps/docs/src/content/docs/queries/aggregate.md
@@ -255,7 +255,10 @@ const topContributors = await store
 
 ## Ordering Aggregated Results
 
-Order by aggregate values:
+Aggregate queries have their own `orderBy(key, direction?)`, called after
+`.aggregate({...})`. `key` is any output name from the fields object — a
+grouped field or an aggregate alias — so `limit()` finally means "top N",
+not "an arbitrary N":
 
 ```typescript
 const topDepartments = await store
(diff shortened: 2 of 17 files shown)


pdlug:
Addressed both review findings:

**P1 (correctness):** confirmed and fixed. The count-aggregate fast path's LIMIT-pushdown guard (`compiler/index.ts`) only checked `ast.orderBy`, not the new `ast.aggregateOrderBy` — so `.optionalTraverse(...).groupByNode(...).aggregate({...}).orderBy("employeeCount", "desc").limit(2)` could push LIMIT into the start CTE before counting, sorting only an arbitrary truncated subset. Extended the guard, and added a cross-backend regression test with companies created in the *reverse* of their expected sort order specifically so a reintroduced pushdown bug can't pass coincidentally (verified: reverted the fix locally, confirmed the new test fails; reapplied, confirmed it passes on both SQLite and real PostgreSQL).

**P2 (docs):** fixed both `aggregate.md` locations and the one in `order.md`'s aggregate section to use the real `.orderBy("key", "desc")` shape instead of the never-implemented `.orderBy((ctx) => ctx.key, "desc")` callback form.

While fixing P2 I found the same phantom callback-based `orderBy((ctx) => ...)` pattern used extensively for *non-aggregate* row queries across `recipes.md`, `temporal.md`, `order.md` (remaining instances), and the `examples/` docs (`product-catalog.md`, `workflow-engine.md`, `audit-trail.md`, `document-management.md` — ~7 files, dozens of instances). That's pre-existing and out of scope for this PR (the real row-query API is `.orderBy(alias, field, direction)`), but worth a follow-up issue since any of those snippets would fail if copy-pasted.

Re-ran the full gate after both fixes: `pnpm typecheck`, `pnpm test` (4553/4553), `pnpm test:postgres` (1837/1837) — all green.

PR #242, opened 2026-07-08
fix: default omitted validFrom to the create timestamp, not open-left NULL

### LINKED ISSUE #240: create without validFrom stores NULL and makes records visible at every asOf instant
## Summary

Creating a node or edge without `validFrom` stores SQL `NULL` in `valid_from`. Temporal filters interpret `NULL` as open-left validity, so records created today appear valid at any historical `asOf` timestamp, including instants before the record existed.

This contradicts existing comments/tests that describe omitted `validFrom` as defaulting to creation time / now.

## Reproduction

Node case:

```ts
const currentCreated = await store.nodes.Author.create({
  name: "Created without validFrom",
});

const ancientRead = await store.nodes.Author.getById(currentCreated.id, {
  temporalMode: "asOf",
  asOf: "1900-01-01T00:00:00.000Z",
});
```

Observed from a local SQLite repro:

```text
node without validFrom visible at 1900 asOf: true
node meta.validFrom: undefined
```

Edge case:

```ts
const edge = await store.edges.knows.create(alice, bob);
const ancientEdges = await store.edges.knows.findFrom(alice, {
  temporalMode: "asOf",
  asOf: "1900-01-01T00:00:00.000Z",
});
```

Observed:

```text
edge meta.validFrom: undefined
edge without validFrom visible at 1900 asOf: true
```

## Root cause

- Node and edge collection create paths only include `validFrom` when the caller passes it.
- Insert SQL writes `sqlNull(params.validFrom)` / `sqlNull(params.validTo)`, so omitted `validFrom` becomes `NULL`.
- Current/asOf filters use `valid_from IS NULL OR valid_from <= asOf`, so `NULL` means valid since forever.

Relevant paths:

- `packages/typegraph/src/store/collections/node-collection.ts`
- `packages/typegraph/src/store/collections/edge-collection.ts`
- `packages/typegraph/src/store/operations/node-operations.ts`
- `packages/typegraph/src/backend/drizzle/operations/nodes.ts`
- `packages/typegraph/src/backend/drizzle/operations/edges.ts`
- `packages/typegraph/src/backend/drizzle/operations/collections.ts`
- `packages/typegraph/src/query/compiler/temporal.ts`

## Expected behavior

If omitted `validFrom` means “valid from creation,” create-like writes should persist the operation timestamp into `valid_from` for both nodes and edges. `asOf` reads before that timestamp should not see the record.

If the intended contract is instead “valid since forever,” the docs and tests should say that explicitly. But current test comments already say omitted `validFrom` defaults to now, so implementation should likely be corrected.

## Test gap

Existing temporal tests can miss this because some fixtures pair the implicit-validFrom edge with a target node whose own `validFrom` is in the future; the target node filter hides the edge bug. Add regression coverage where both endpoints are valid at the historical `asOf`, but the edge itself was created without `validFrom` later.

## Notes

This was surfaced while reviewing PR #232 example 26.

pdlug:
Creating a node or edge without an explicit `validFrom` stored SQL `NULL` in `valid_from`. Temporal filters interpret `NULL` as open-left validity, so records created today were visible at any historical `asOf` instant — including instants before the record existed. This contradicted the documented "omitted `validFrom` defaults to now" contract.

## Root cause & fix

Every node/edge insert path (single, no-return, batch, batch-returning — for both the store's collection API and the interchange bulk-import path) funnels through 8 SQL-builder call sites in `backend/drizzle/operations/{nodes,edges}.ts`. Each already threads a `timestamp` parameter (the same one stamped into `created_at`/`updated_at`) but discarded it for `validFrom` when omitted (`sqlNull(params.validFrom)` → `NULL`).

Fixed at that single choke point with a new `resolveValidFrom(validFrom, timestamp)` helper in `shared.ts` (mirroring the existing `sqlNull` helper): an omitted `validFrom` now resolves to the exact same timestamp as `created_at`, applying uniformly across every write path with no changes needed to the store-operation layer, `import.ts`, or the temporal query compiler (`valid_from IS NULL` handling stays, correctly, for any pre-existing legacy-NULL rows).

## Second-order fix: `branch()` working-copy fidelity

A code review caught a real regression this change introduced: `branch()`'s default working-copy strategy clones a store via `exportGraph`/`importGraph` without `includeTemporal: true`, so `validFrom` was dropped on export. Post-fix, re-importing an omitted `validFrom` re-stamps it to the *clone's own* (later) creation time instead of the base row's original instant — silently narrowing a fork's valid-time window relative to its base for any `asOf` between the row's real creation and the fork.

Fixed by passing `includeTemporal: true` in `cloneWorkingCopyStrategy`'s export call, so a fork's `validFrom`/`validTo` now match the base exactly (verified safe against `base@V` fingerprinting, which is computed only off the original store, never the clone). Added a dedicated regression test (`branch.test.ts`) that fails without this fix and passes with it, on both SQLite and PGlite backends.

## Also fixed (review findings)

- Removed a new test assertion that looked like it validated the fix but was actually driven by an unrelated `deleted_at` exclusion (false-confidence regression coverage).
- Updated `limitations.md`, `interchange.md`, and `schemas-stores.md` to reflect the new default behavior and the `exportGraph`/`importGraph` round-trip caveat.
- Added `validFrom`-defaults-to-creation-time JSDoc to every sibling collection method that accepts the option (`createFromRecord`, `upsertById`, `bulkCreate`, `bulkInsert`, `bulkUpsertById`, and edge equivalents).
- Added a changeset (`patch`).

Fixes #240

### DIFF
--- a/.changeset/validfrom-defaults-to-creation-time.md
+++ b/.changeset/validfrom-defaults-to-creation-time.md
@@ -1,0 +1,25 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Fix: creating a node or edge without an explicit `validFrom` now stamps the
+operation's own creation timestamp instead of storing SQL `NULL`.
+
+`NULL` is interpreted by temporal filters as open-left validity ("valid
+since forever"), so a record created without `validFrom` was visible at
+*any* historical `asOf` instant — including ones before the record existed.
+This contradicted the documented contract ("omitted `validFrom` defaults to
+now") and is fixed at the insert layer for every write path: `create`,
+`createFromRecord`, `upsertById`/`upsertByIdFromRecord` (create branch),
+`bulkCreate`, `bulkInsert`, `bulkUpsertById`, and get-or-create, for both
+nodes and edges.
+
+`branch()`'s working-copy clone now also exports with `includeTemporal:
+true`, so a fork's `validFrom`/`validTo` exactly match the base's — without
+this, the clone would re-stamp any implicit `validFrom` to the fork's own
+(later) creation time, narrowing the fork's valid-time window relative to
+the base it was cloned from.
+
+`exportGraph`/`importGraph` round trips still default `includeTemporal` to
+`false`; without it, imported records get a fresh `validFrom` at import
+time rather than the source's original value (see the Interchange docs).
--- a/.changeset/validfrom-defaults-to-creation-time.md
+++ b/.changeset/validfrom-defaults-to-creation-time.md
@@ -18,7 +18,12 @@ nodes and edges.
 true`, so a fork's `validFrom`/`validTo` exactly match the base's — without
 this, the clone would re-stamp any implicit `validFrom` to the fork's own
 (later) creation time, narrowing the fork's valid-time window relative to
-the base it was cloned from.
+the base it was cloned from. This includes rows that still have a `NULL`
+`valid_from` (predating this fix, or written directly via the backend):
+`exportGraph`/`importGraph` now round-trip a confirmed open-left window as
+an explicit `null` rather than silently dropping it, so a legacy row's
+"valid since forever" semantics survive a clone unchanged instead of being
+narrowed to the clone's own creation time.
 
 `exportGraph`/`importGraph` round trips still default `includeTemporal` to
 `false`; without it, imported records get a fresh `validFrom` at import
--- a/apps/docs/src/content/docs/core-concepts.md
+++ b/apps/docs/src/content/docs/core-concepts.md
@@ -122,7 +122,7 @@ const alice = await store.nodes.Person.create({ name: "Alice", email: "a@example
 //     createdAt: "2024-01-15T10:30:00.000Z",
 //     updatedAt: "2024-01-15T10:30:00.000Z",
 //     deletedAt: undefined,
-//     validFrom: undefined,
+//     validFrom: "2024-01-15T10:30:00.000Z", // defaults to createdAt when omitted
 //     validTo: undefined,
 //   }
 // }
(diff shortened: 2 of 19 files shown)


pdlug:
Addressed the P1 finding: legacy rows with `valid_from = NULL` (predating this fix, or written directly via the backend) were losing their open-left fidelity on a `branch()` clone.

**Root cause confirmed**: `exportGraph(includeTemporal: true)` only emitted `validFrom` when `row.valid_from` was truthy, so a NULL row's `validFrom` key was omitted from the wire payload — indistinguishable from "not requested". On import, an omitted `validFrom` now correctly defaults to the import's own timestamp (that's the whole point of #240), which silently narrowed a legacy "valid since forever" row to "valid since clone time".

**Fix** (f1521428): gave the interchange wire format a real three-state `validFrom` — omitted (not requested, defaults on import) vs. explicit `null` (confirmed open-left, preserved as NULL) vs. a string (explicit value). `InsertNodeParams`/`InsertEdgeParams.validFrom` widen to `string | null | undefined` to carry the "preserve as NULL" signal down to the SQL layer. The public collection API (`store.nodes.X.create()`, etc.) is unaffected — it still only produces `string | undefined`, so ordinary callers can't produce this state; only interchange import can, by design.

**New regression coverage**:
- `tests/interchange.test.ts`: exports/imports a directly-inserted NULL row, asserts it round-trips as explicit `null` (not a re-stamped timestamp).
- `tests/graph-merge/branch.test.ts`: reproduces your exact scenario — a legacy NULL-`valid_from` row stays visible at an ancient `asOf` on the fork, matching the base.

Both confirmed to fail without the fix (reverted locally to verify) and pass with it, on SQLite + PGlite. Full suites green: SQLite 4627/4627, Postgres 1849/1849.

Issue #255, opened 2026-07-14
withRecordedTransaction should return TransactionOutcome<T>: exactly-once cursors and receipt-anchored replay do not compose


pdlug:
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) — the first real consumer of the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide — against 0.35.0.

**This is the highest-value item from that integration, and the smallest change on the list.** It makes the guide's own advice implementable; today two of its sections are mutually exclusive.

## The bind

An event-log materializer wants two things at once, and the guide recommends both:

1. **Exactly-once cursors** — *"Write cursors within the same `store.transaction` callback for exactly-once materialization"* (Cursor Management).
2. **Receipt-anchored replay** — *"the receipt's write count indicates completed intents, while the `recorded` anchor enables replay reconstruction at specific source offsets"* (Transaction Receipts).

You can have either, not both:

| path | cursor in the transaction? | receipt? |
| --- | --- | --- |
| `transactionWithReceipt(fn)` | ❌ owns its own boundary; `tx.sql` is guarded under `history: true` (#252) | ✅ |
| `withRecordedTransaction(externalTx, fn)` | ✅ caller owns the boundary | ❌ returns `T`, not `TransactionOutcome<T>` |

So the exactly-once consumer gets **no receipt**, and therefore:

- **no `writes.total`** → it cannot detect a dropped change (a projector that wrote to the wrong store, swallowed a write error, or handled no shape). The cursor advances past the change and it is lost forever.
- **no per-transaction `recorded`** → it has no replay anchor. The only anchor left is `store.recordedNow()`, which is the **graph-global** high-water: under any concurrent writer it returns *someone else's* commit instant, and `asOfRecorded(thatInstant)` reconstructs a belief this stream never produced. That is precisely the value `receipt.recorded` was added to replace.

The remaining workaround — modelling the cursor as a node in the projected graph — is worse: the cursor's own write lands in `receipt.writes.total`, so a dropped change becomes undetectable (`total === 1` from the cursor alone). Filed separately as the scoped-receipt request.

## Verified

`withRecordedTransaction` genuinely delivers the atomicity (better-sqlite3, `history: true`): a projector that throws leaves the cursor unadvanced and the node absent. It just hands back `fn`'s return value and nothing else.

```ts
await db.run(sql`BEGIN IMMEDIATE`);
[...]

PR #260, opened 2026-07-14
fix: make the store transaction surface honest (#252, #253, #254, #258)

### LINKED ISSUE #252: withRecordedTransaction's JSDoc promises `tx.sql`, but it is `never` in the type and throws at runtime
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide on 0.35.0.

## The contradiction

`Store.withRecordedTransaction`'s JSDoc says:

> Like `Store.withTransaction`, the graph writes and any raw `tx.sql` statements share the caller's one pinned connection. TypeGraph serializes the statements its collections issue; sequence your own raw statements (and don't `Promise.all` them with graph writes) so two queries never race on that connection.

But `withRecordedTransaction`'s history overload hands back a `HistoryTransactionContext<G>`, which types `sql` away entirely:

```ts
type HistoryTransactionContext<G extends GraphDef> = Omit<TransactionContext<G>, "backend" | "sql"> & Readonly<{
    backend: HistorySafeTransactionBackend;
    sql?: never;          // <-- typed away
}>;
```

…and at runtime it is present-but-throwing:

```
ConfigurationError: tx.sql is not available when history capture is enabled
because raw SQL writes bypass recorded-time capture.
```

So the one method that exists specifically to let a history-enabled store join a caller-owned transaction documents an escape hatch that its own type forbids and its own runtime rejects.

## Why it matters

`withRecordedTransaction` is the *only* way to commit graph writes atomically with your own relational rows on a `history: true` store (`store.transaction`'s `tx.sql` is guarded, and `withTransaction` throws — see the sibling issues). A reader following the JSDoc reaches for `tx.sql` to write, say, a stream-cursor row inside the transaction, and hits a dead end on the exact path the method exists to serve.

## Repro

```ts
const { backend, db } = createLocalSqliteBackend();
const [store] = await createStoreWithSchema(graph, backend, { history: true });

await db.run(sql`BEGIN IMMEDIATE`);
await store.withRecordedTransaction(db, async (tx) => {
  console.log(typeof (tx as any).sql);   // "function"/object — not undefined, despite `sql?: never`
  await (tx as any).sql.run(sql`SELECT 1`);
  // ConfigurationError: tx.sql is not available when history capture is enabled…
});
```

## Suggested fix

The correct guidance is the inverse of what the JSDoc says, and it costs nothing: **on the adopted path the caller already holds the transaction handle** (they passed it in as `externalTx`), so they should write their own tables through *that*, not through `tx.sql`. Verified working on better-sqlite3 (manual `BEGIN IMMEDIATE`/`COMMIT`, since its sync driver rejects an async `db.transaction()` callback), and it generalizes to Postgres via `db.transaction(async (pgTx) => { await store.withRecordedTransaction(pgTx, …); await pgTx.insert(cursor)… })`.

Proposed JSDoc:

> Write your own relational tables through the external transaction handle you passed in — it *is* the pinned connection. `tx.sql` is unavailable under history capture (raw SQL bypasses recorded-time capture); this is the sanctioned way to get cross-store atomicity on a history-enabled store.

### LINKED ISSUE #253: withTransaction() on a HistoryStore typechecks but always throws — should be a compile error
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide on 0.35.0. Sibling of #252.

## Problem

`store.withTransaction(externalTx)` on a history-enabled store **typechecks cleanly** and then throws at runtime:

```
ConfigurationError: withTransaction() has no recorded-time capture flush point when history is enabled.
```

This is fully preventable at compile time. The type system already knows the store is a `HistoryStore<G>`:

- `createStoreWithSchema(graph, backend, { history: true })` has a dedicated overload returning `Promise<[HistoryStore<G>, …]>`.
- Sibling methods already use exactly this discrimination — `transactionWithReceipt` and `withRecordedTransaction` both declare `this: HistoryStore<G>` overloads.

`withTransaction` just lacks the negative constraint, so the one method that *can't* work on a history store is the one method that doesn't say so statically.

## Repro (typechecks with `tsc --strict`, throws at runtime)

```ts
const [history] = await createStoreWithSchema(graph, backend, { history: true });
const adopted = history.withTransaction(db);   // ✅ tsc is happy
                                               // 💥 ConfigurationError at runtime
```

## Suggested fix

Add a negative overload so it fails at the call site, and point at the working alternative:

```ts
/** @deprecated Not available on a history-enabled store — use {@link Store.withRecordedTransaction}. */
withTransaction(this: HistoryStore<G>, externalTx: never): never;
withTransaction(externalTx: AdoptedTransaction): TransactionContext<G>;
```

Any shape that turns this into a compile error works; the value is that the error arrives at the call site with `withRecordedTransaction` named in it, rather than in production.

### LINKED ISSUE #254: tx.sql is typed `sql?: never` but is present-and-throwing at runtime — feature detection is broken
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide on 0.35.0. Sibling of #252 / #253.

## Problem

On a history-enabled store, `tx.sql` disagrees with itself in **both** directions:

- **The type says it's absent.** `HistoryTransactionContext` declares `sql?: never`, so `undefined` is the only assignable value.
- **The runtime says it's present.** `typeof tx.sql !== "undefined"` — it's a throwing guard object, not `undefined`.

The documented contract for `tx.sql` says it "is `undefined` only on the non-transactional fallback (`backend.capabilities.transactions === false` — e.g. Cloudflare D1, `drizzle-orm/neon-http`)". So `undefined` is a *meaningful, documented signal* about backend capability. A history store now violates that: it is fully transactional, yet its `tx.sql` is neither usable nor absent.

## Consequence

The documented feature-detection pattern silently breaks. This reads as truthy and then throws:

```ts
await store.transaction(async (tx) => {
  if (tx.sql) {                       // truthy on a history store…
    await tx.sql.run(sql`…`);         // 💥 ConfigurationError
  } else {
    await fallbackWithoutAtomicity(); // never reached
  }
});
```

A caller writing backend-portable code (D1/neon-http fallback vs. transactional path) cannot distinguish "no transaction support" from "history capture forbids raw SQL" without catching an exception and string-matching the message.

## Repro

```ts
const [store] = await createStoreWithSchema(graph, backend, { history: true });
await store.transaction(async (tx) => {
  console.log(typeof tx.sql);  // not "undefined" — but the type insists it is
});
```

## Suggested fix

Pick one and make type and runtime agree:

1. **Make it genuinely `undefined`** at runtime on history stores, matching `sql?: never`. Feature detection then works, but the failure mode for someone who ignores the type is `undefined is not a function` — a worse error than today's.
2. **Keep the fail-loud guard** (better DX for the mistake it catches) and *type it as present*, documenting that it throws under history capture. Then `if (tx.sql)` is honestly true, and the guard's message does the teaching.

I'd lean toward (2) — the fail-loud message is genuinely good and worth keeping — plus a separate, honest capability signal for the portability case (e.g. `tx.canWriteRawSql`, or reusing `backend.capabilities`), so the D1/neon-http fallback branch has something real to test. Related: a branchable error code for these guards is filed separately.

### LINKED ISSUE #258: Branchable error code for the recorded-capture guards (currently only distinguishable by message string)
Both recorded-capture guards throw a bare `ConfigurationError`:

- `tx.sql is not available when history capture is enabled because raw SQL writes bypass recorded-time capture.`
- `withTransaction() has no recorded-time capture flush point when history is enabled.`

A caller who wants to handle these — e.g. a library that supports both history and non-history stores and picks its transaction strategy accordingly, which is exactly what a materializer that wants to work on either kind of belief store must do — can only tell them apart by **substring-matching the message**. That's brittle across releases and is the kind of thing that breaks silently on a message reword.

0.35 did great work here for the fixed-shape errors: every `TypeGraphError` subclass with a fixed-shape `details` payload now declares a narrowed `readonly details` type (`RestrictedDeleteErrorDetails`, `UniquenessErrorDetails`, …). But `ConfigurationError` is explicitly in the "intentionally open, per-call-site details" bucket and was left unchanged — so these two guards have nothing to branch on.

The internal name `RECORDED_CAPTURE_REQUIRES_CALLBACK_TRANSACTION` is even referenced in the `withTransaction` JSDoc, so a stable identifier already exists conceptually; it just isn't reachable from the caught error.

## Ask

Give these guards something programmatic to branch on. Any of:

1. A `code` field on `ConfigurationError` (`"RECORDED_CAPTURE_REQUIRES_CALLBACK_TRANSACTION"`, `"RAW_SQL_BYPASSES_RECORDED_CAPTURE"`), populated at these throw sites.
2. Dedicated subclasses (`RecordedCaptureConfigurationError`) that `instanceof` can catch.
3. A narrowed `details` payload for just these sites, following the 0.35 pattern.

(1) is the smallest and matches the fact that the codes already exist by name internally.

## Related

If #254 is resolved by keeping `tx.sql` as a fail-loud guard rather than making it `undefined`, then an error code becomes the *only* way a portable caller can distinguish "this backend has no transactions" (D1, neon-http) from "history capture forbids raw SQL here" — so these two are worth deciding together.

pdlug:
Fixes #252, fixes #253, fixes #254, closes #258.

Four API-surface honesty fixes on the store transaction surface, found by the first real consumer (`@nicia-ai/agent-stream-graph`). One PR because they are one story — the same ~100 lines of surface, decided together per the cross-references in the issues.

## #252 — `withRecordedTransaction` JSDoc promised `tx.sql`
The JSDoc now documents the inverse (and correct) contract: **write your own relational tables through the external transaction handle you passed in** — it *is* the pinned connection — with a worked Postgres/libsql example. `tx.sql` is unavailable under history capture because raw SQL bypasses recorded-time capture.

## #253 — `withTransaction()` on a `HistoryStore` is now a compile error
The issue's suggested overload shape (`externalTx: never`) doesn't actually fire — overload resolution falls through to the general signature, and `historyEnabled` is a `boolean` getter on the base `Store` class, so a `this`-constraint can't discriminate either. Implemented instead as a single generic signature with a conditional parameter type:

```ts
withTransaction<This extends Store<G>>(
  this: This,
  externalTx: This extends HistoryStore<G> ? HistoryWithTransactionUnavailable : AdoptedTransaction,
): This extends HistoryStore<G> ? never : TransactionContext<G>;
```

`HistoryWithTransactionUnavailable` is a descriptive string-literal type, so the call-site error itself names `withRecordedTransaction`. Plain and generic `Store<G>` call sites compile unchanged (pinned by type tests); the runtime guard remains for suppressed/untyped callers.

## #254 — `tx.sql` feature detection
Decision: **keep the fail-loud guard** (its teaching error is worth having) and **keep `sql?: never`** on `HistoryTransactionContext` — changing the type to "present" would break `HistoryTransactionContext → TransactionContext` assignability, which projector helpers rely on. Instead, feature detection gets an honest, explicit signal: every transaction context now carries

```ts
sqlAvailability: "available" | "history" | "revisionTracking" | "unavailable"
```

- `"available"` — `tx.sql` is the usable raw handle
- `"history"` / `"revisionTracking"` — transaction is real, but `tx.sql` is a fail-loud guard (raw SQL would bypass capture / the revision anchor)
- `"unavailable"` — non-transactional fallback (D1, neon-http); `tx.sql` is `undefined`

`tx.sql === undefined` now means exactly what the docs always claimed: the non-transactional fallback, nothing else. This also covers the revision-tracking case, which is a runtime store option and cannot be discriminated statically.

## #258 — branchable error codes
Discovery: the codes **already ship** in 0.35 — both guards have carried `details.code` since #186 (`RECORDED_CAPTURE_REQUIRES_CALLBACK_TRANSACTION`, `RECORDED_CAPTURE_RAW_SQL_DISABLED`, plus `REVISION_TRACKING_RAW_SQL_DISABLED`). So this formalizes rather than invents: exported `RECORDED_CAPTURE_GUARD_CODES` / `RecordedCaptureGuardCode` (shipped values verbatim), the `code` now set inside the error factory so no throw site can omit it (one call site in `recorded-capture.ts` previously did), and a typed predicate `isRecordedCaptureGuardError(error, code?)` following the `StoreNotInitializedReason` precedent. Combined with `sqlAvailability`, a portable caller can now distinguish "this backend has no transactions" from "history capture forbids raw SQL here" without string-matching messages.

## Tests
New `tests/transaction-surface-honesty.test.ts`: all four `sqlAvailability` states (`tx.sql` `undefined` only in `"unavailable"`), exact `details.code` on all three guards, `isRecordedCaptureGuardError` narrowing (including rejecting a code-less `ConfigurationError`), and `@ts-expect-error` type tests for #253 in both directions.

Gates: `pnpm fix` clean, typecheck clean, SQLite 4861 passed, PostgreSQL 1947 passed.

Note: a follow-up PR (#255/#257 — receipts on the adopted path + scoped measurement) is stacked on this branch.


pdlug:
Review fixes pushed in 02c1694b: (1) `RecordedCaptureGuardError<C>` is now generic with a code-specific `isRecordedCaptureGuardError` overload, so passing a code narrows `details.code` to that literal (type test added); (2) the guard examples in the JSDoc and errors.md now widen through `const store: Store<G> = historyStore` so they compile — the direct call is the compile error this PR introduces. CI reruns on the new commit (previous red was the unrelated libsql vector timeouts).

PR #261, opened 2026-07-14
feat: return a receipt from withRecordedTransaction, add scoped tx.measure (#255, #257)

### LINKED ISSUE #255: withRecordedTransaction should return TransactionOutcome<T>: exactly-once cursors and receipt-anchored replay do not compose

### LINKED ISSUE #257: Scoped receipts: a consumer's own bookkeeping writes contaminate receipt.writes.total
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide on 0.35.0. Related: #255.

## Problem

`receipt.writes` counts every completed write intent **in the transaction**, with no attribution. So a framework that runs *user code* inside a transaction cannot separate "what the user's code wrote" from "what I wrote".

That is exactly the shape of an event-log materializer: it invokes a user-supplied `project(tx, change)` callback and needs to know whether *the projector* wrote — because a non-`delete` change that writes nothing is a **dropped change**, and advancing the cursor past it loses the event permanently.

The guide recommends storing cursors as graph nodes. Do that in the same transaction and the two collide:

```ts
await store.transactionWithReceipt(async (tx) => {
  await project(tx, change);                                  // drops the change — writes nothing
  await tx.nodes.Cursor.upsertById("s1", { offset: "002" });  // the consumer's own bookkeeping
});
// receipt.writes: {"nodes":{"Cursor":1},"edges":{},"total":1}
```

`writes.total > 0`, so the dropped change is undetectable, the cursor advances past it, and the event is gone. The guide recommends both halves of this trap in adjacent sections.

Our workaround is to keep the cursor in a **separate store** entirely, which costs us the exactly-once property (#255) and forces at-least-once + idempotence. That's a fine trade, but it's forced by a receipt limitation rather than chosen.

## Ask

A way to scope a receipt to a region of the transaction — measure what a callback wrote, without the surrounding bookkeeping:

```ts
await store.transactionWithReceipt(async (tx) => {
  const projected = await tx.measure(() => project(tx, change));  // sub-receipt
  if (projected.writes.total === 0 && change.operation !== "delete") {
    throw new DroppedChangeError(change);
  }
  await tx.nodes.Cursor.upsertById("s1", { offset: change.offset });  // not counted in `projected`
});
```

Shape is open — `tx.measure(fn)` returning a `TransactionReceipt`-alike, or a counter snapshot (`tx.writeCount()` before/after) that a caller can diff. The counters already exist per-transaction; this only needs a way to read them mid-flight or scope them to a callback.

This would let a materializer checkpoint **in-graph, in the same transaction** (exactly-once) *and* still detect a dropped change — the combination that is impossible today. It composes with #255 rather than duplicating it: #255 gives the adopted path a receipt at all; this makes that receipt attributable when the consumer also writes.

## Context: why the write-count signal matters so much here

We previously hand-rolled this by wrapping `tx` in a `Proxy` that flagged when any write method on `tx.nodes`/`tx.edges` resolved, with a hand-synced list of write-method names kept honest by a `satisfies Record<keyof NodeWrites | keyof EdgeWrites, true>`. `transactionWithReceipt` let us delete all ~110 lines of that in 0.35 and matched its semantics rule-for-rule (bulk counts by input length, a rejected write counts 0, a `delete` of an absent id counts 1 on resolve) — a genuinely great API. Scoping is the one thing it can't express, and it's the thing that keeps the cursor out of the transaction.

pdlug:
Closes #255, closes #257. **Stacked on #260** (`fix/transaction-surface-honesty`) — merge that first; GitHub will retarget this to `main`.

These two land together because they are the same wound from two sides (as argued in the issues): #255 gives the adopted path a receipt at all; #257 makes that receipt attributable when the consumer also writes. Together they enable what the event-log guide describes but 0.35 cannot do: **exactly-once in-graph cursors + drop detection + per-offset replay anchors in one transaction.**

## #255 — `withRecordedTransaction` now returns `TransactionOutcome<T>` (BREAKING)

```ts
const { result, receipt } = await store.withRecordedTransaction(externalTx, async (tx) => { ... });
receipt.writes.total; // completed write intents — drop detection
receipt.recorded;     // this transaction's recorded commit instant — the replay anchor
```

The data was already there — the capture scope's `flush()` returns the per-graph recorded instants; they were being discarded. The non-history overload also returns an outcome (writes counted, `recorded` always `undefined`), via a uniform empty-flush shape. Post-flush sealed-session behavior is unchanged.

Deliberate breaking change for 0.36 rather than a parallel `*WithReceipt` method: one adopted-path method, one return shape. Migration is one destructure (changeset carries it). No non-test callers existed in-repo.

## #257 — `tx.measure(fn)` scoped receipts

```ts
await store.transactionWithReceipt(async (tx) => {
  const projected = await tx.measure(async () => project(tx, change));
  if (projected.receipt.writes.total === 0 && change.operation !== "delete") {
    throw new DroppedChangeError(change); // cursor write below can no longer mask this
  }
  await tx.nodes.Cursor.upsertById("s1", { offset: change.offset }); // outer receipt only
});
```

- Available exactly where a recorder exists: `transactionWithReceipt` and `withRecordedTransaction` — typed statically via new `MeasurableTransactionContext<G>` / `MeasurableHistoryTransactionContext<G>` (assignability chain preserved: history-measurable ⊑ measurable ⊑ `TransactionContext`, so projector helpers keep compiling). Plain `transaction()` contexts have no `measure`, at runtime and in the types.
- Counting inherits the receipt rules verbatim (count on resolve, intent pinned at call, bulk by input length, rejected counts 0). A write counts in a scope iff it resolves while the scope is open — await your writes inside the measured callback. Nested/overlapping scopes count independently. Scoped receipts never carry `recorded` (that is a per-transaction flush concern).
- Writes inside a scope still count in the outer receipt — a scope is a view, not a partition.

## Tests
- Adopted path: history receipt + recorded anchor, read-only → `recorded` undefined, non-history overload, and the full composed materializer scenario (measured projector + out-of-scope cursor + dropped-change detection) on a real adopted better-sqlite3 transaction.
- `measure`: scope-vs-outer counts, writes after close, nesting/overlap, rejected-fn propagation, bulk-by-length, `recorded` undefined — in the shared cross-backend integration suite (runs on SQLite and PostgreSQL).
- PG adopted-path receipt + anchor asserted in the advisory-lock suite.
- Type-level: `test-d` pins the new return shape, `measure` presence/absence, and the assignability chain.


pdlug:
Review fixes pushed in 9daa8bb1 (+ base merge f. 02c1694b): **P1 overlapping-measure cross-count** — `tx.measure` redesigned to scoped-context attribution: the callback receives a scoped context and the scope counts exactly the writes made through it, so concurrent/overlapping measures are cross-count-safe by construction (concurrent-sibling regression test added; dynamic `getNodeCollection` writes are also scope-attributed). **P1 post-write seal** — `session.assertOpen()` now runs at the top of all 16 recorded-overlay write methods; an escaped-context write fails before the live mutation (test: swallowed rejection + COMMIT → row not persisted). **P2 non-history seal** — the non-history `withRecordedTransaction` path seals its context when `fn` resolves; post-return writes throw instead of committing uncounted. Gates: SQLite 4895 / PG 1971 green post-merge.

PR #262, opened 2026-07-14
feat: add opt-in coalesceUnchangedUpserts store option (#256)

### LINKED ISSUE #256: Coalesce value-identical upserts (or give the receipt rows-affected fidelity): replay rewrites every row and churns recorded history
Found while implementing [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against the [Materializing External Event Logs](https://typegraph.dev/materializing-event-logs/) guide on 0.35.0.

This is the biggest **practical** gap for event-log materialization — the one you feel in production rather than at the type level.

## Problem

An at-least-once stream re-delivers, and a rebuild replays the whole log. The guide's own framing calls this out as a core requirement:

> **At-least-once delivery** — changes may be delivered multiple times after crashes
> **Replay capability** — reprocessing historical events for recovery or rebuilds

Idempotent projectors converge the *live* state correctly — `upsertById` gets the right answer. But every re-delivery of a **byte-identical** value still performs a real write: `upsertById` on an existing id calls `updateNode` unconditionally (no dirty check), so it allocates a fresh recorded instant and captures a new history row.

```
first  : writes=1  recorded=2026-07-14T16:36:40.756Z
replay : writes=1  recorded=2026-07-14T16:36:40.757Z   <- identical props
replay2: writes=1  recorded=2026-07-14T16:36:40.758Z   <- identical props
```

Two consequences:

1. **A full replay of an N-event log rewrites every row and grows recorded history by N**, even though the graph converged to an identical state. Recovery/rebuild — the workload the guide explicitly recommends — is the one that inflates history most. Recorded time then shows a dense band of "changes" that changed nothing, which is exactly what a bitemporal audit trail should not say.
2. **`receipt.recorded` cannot distinguish "converged, nothing changed" from "real change."** For an at-least-once consumer that is *the* question, and it currently has no answer.

## Ask

Either of these fixes it; the first is the one I'd want:

1. **Coalesce value-identical upserts** — a `coalesceUnchangedUpserts` store option (I gather this has been floated before; this use case is the concrete argument for it). A dirty-check on the props before `updateNode` skips the write and the capture.
2. **Rows-affected fidelity in the receipt** — explicitly out of scope for v1 per the `TransactionReceipt` docs ("a future extension could ask backends to return row counts"). That would let the consumer decide, at the cost of still doing the write.

## The receipt design already accommodates option 1

Worth noting, because it means this composes cleanly rather than needing new signals: a coalesced no-op upsert would surface as

- `writes.total === 1` — the write intent *did* complete, so dropped-change detection is untouched, and
- `recorded === undefined` — nothing was captured.

That is the **identical shape as a no-op delete** (deleting an absent key), which the receipt already produces and which our consumer already handles by carrying the prior anchor forward to that offset. So `coalesceUnchangedUpserts` needs no new receipt surface at all — it just makes an existing, already-handled case fire more often. The two-signal split (`writes.total` vs `recorded`) is what makes it safe.

Opt-in is right: callers who *want* an audit row per re-delivery (proof the event was reprocessed) should keep today's behavior.

pdlug:
Closes #256.

Opt-in store option for at-least-once / replay materializers: with `coalesceUnchangedUpserts: true`, an upsert whose validated props are value-identical to the existing live row performs **no write at all** — no `updateNode`, no recorded-history row, no revision-clock advance, no operation hooks — and resolves with the existing node. A full N-event replay of an already-converged log now writes nothing and leaves recorded time untouched, instead of rewriting every row and growing history by N.

Default is **off**: some consumers want an audit row per re-delivery (proof the event was reprocessed). A per-call flag was considered and rejected — #256 asks for a store option, and the workloads that want this (replay materializers, merge) want it store-wide.

## Coalescing rules (all must hold, else the normal write happens)
1. A live row exists for the id (else it's a create).
2. The row is not soft-deleted (upsert resurrects; never coalesced).
3. No explicit `validFrom`/`validTo` was passed (an explicit temporal override is a real request).
4. Validated new props deep-equal the stored props (`canonicalEqual` — key-order-independent, schema defaults applied), compared on exactly what the write path would write: the dirty check shares `resolveNodeUpdateProps`/`resolveEdgeUpdateProps` with the update path, so the two cannot disagree.

Covers the full upsert-by-id surface: node `upsertById` / `upsertByIdFromRecord` / `bulkUpsertById`, edge `bulkUpsertById` (edges have no single upsert; endpoints are never rewritten by upsert, so props-only comparison is exact). Bulk batches coalesce per item with input-order-preserved results. The option reaches transaction-scoped collections automatically.

## Receipt semantics (verified — the composability claim in #256 holds)
A coalesced replay through `transactionWithReceipt` surfaces as `writes.total === 1` with `recorded === undefined` — byte-identical to the no-op delete shape consumers already handle by carrying the prior anchor forward. No new receipt surface needed; the two-signal split does the work. (Caveat inherited from receipt granularity: the shape is per-transaction, so a coalesced upsert mixed with other real writes in one transaction is masked — same as no-op deletes today.)

## graph-merge interaction: sound, no bypass
Merge commits through tx collections and inherits the flag. Safe because: the base@V guard runs before any upsert and reads the revision clock, unaffected by elided writes; conflicts are resolved at plan time and a coalesced write is by definition byte-identical; the final merged state is identical — unchanged rows just don't get re-stamped with a new recorded instant, which is the option's purpose. Full merge suites (482 tests) green. Documented on `commitPlan`.

## Honest cost note
With the flag ON, a genuinely-*changed* upsert pays one extra Zod validation (the dirty check); the write path deliberately re-resolves under its own transaction (re-read under lock — threading the pre-lock resolution in would be a TOCTOU). The replay path this targets validates once and skips the write entirely.

## Tests
- Shared cross-backend suite (`tests/backends/integration/coalesce-upserts.ts`, SQLite + PostgreSQL): the full rules matrix (unchanged/changed/soft-deleted/explicit-validFrom/flag-off), key-order independence, mixed bulk batches, edge symmetry, history assertions (no new recorded row, `recordedNow()` stable, capture resumes on real change), revision-anchor non-advance, and the #256 receipt shape end-to-end.
- Property test: arbitrary props re-upserted identically is history-idempotent under the flag.

Gates: typecheck clean, SQLite 4888 passed, PostgreSQL 1983 passed, `pnpm fix` clean (the docs-app lint failure seen mid-review was a fresh-worktree Astro codegen artifact, resolved by `astro sync`; nothing in this diff).

Independent of #260/#261 (different seams); all three target the same release.

### DIFF
--- a/.changeset/coalesce-unchanged-upserts.md
+++ b/.changeset/coalesce-unchanged-upserts.md
@@ -1,0 +1,34 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Add an opt-in `coalesceUnchangedUpserts` store option for at-least-once /
+replay materializers.
+
+Idempotent event-log projectors converge live state correctly, but every
+re-delivery of a byte-identical value still performed a real write:
+`upsertById` on an existing id called `updateNode` unconditionally, allocating
+a fresh recorded instant and a new history row. A full replay of an N-event log
+therefore rewrote every row and grew recorded history by N — the recovery /
+rebuild workload inflates history the most.
+
+With `createStore(graph, backend, { coalesceUnchangedUpserts: true })`, an
+`upsertById` (or `bulkUpsertById` item) whose validated props are
+value-identical to the existing **live** row performs **no write at all**: no
+`updateNode`, no recorded-time capture, no history row, no revision-anchor
+advance, and no `update` operation hooks. It resolves with the existing node.
+The dirty-check compares the storage-normalized representation (props run
+through the kind's Zod schema, key-order-independent), so it answers exactly
+"would the persisted value differ?".
+
+A write still happens (never coalesced) when the row is soft-deleted (an upsert
+resurrects it), when an explicit `validFrom` / `validTo` is passed, or when any
+prop differs. Default off, because some consumers want an audit row per
+re-delivery. Covered symmetrically for edge `bulkUpsertById` (props only —
+endpoints are the edge's identity).
+
+Receipt semantics are unchanged and need no new signal: a coalesced upsert
+still counts as one write intent (`writes.total`) but captures nothing
+(`recorded` stays `undefined`) — the same two-signal shape as a no-op delete,
+which at-least-once consumers already handle by carrying the prior anchor
+forward.
--- a/apps/docs/src/content/docs/schemas-stores.md
+++ b/apps/docs/src/content/docs/schemas-stores.md
@@ -460,6 +460,7 @@ function createStore<G extends GraphDef>(
 | `schema` | `SqlSchema` | Custom table name configuration created with `createSqlSchema(...)` |
 | `queryDefaults.traversalExpansion` | `TraversalExpansion` | Default ontology expansion mode for traversals (default: `"inverse"`) |
 | `autoRefreshStatistics` | `false \| number` | Row threshold at which a single autocommit `bulkCreate`/`bulkInsert` triggers an automatic planner-statistics refresh (default: `1000`); `false` disables. See [Refreshing planner statistics](/backend-setup#refreshing-planner-statistics-after-bulk-loads). |
+| `coalesceUnchangedUpserts` | `boolean` | Skip the write for an `upsertById` / `bulkUpsertById` item whose validated props already equal the existing live row (default: `false`). For at-least-once / replay materializers: a byte-identical re-delivery performs no write, no history row, and no revision advance. See [`upsertById`](#upsertbyidid-props-options) and [Materializing external event logs](/materializing-event-logs). |
 
 **Example:**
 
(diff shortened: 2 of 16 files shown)


pdlug:
Review fixes pushed in df076cc0: **P1 duplicate-id bulk upserts** — items are now dirty-checked against batch-local pending state (last non-coalesced value per id), restoring last-write-wins (exact probe from the review is a regression test, nodes + edges, cross-backend); **P2 onError bypass** — a validation failure in the dirty check falls through to the hooked write path, so `onError` fires exactly as with the flag off; **P2 stats refresh** — `maybeRefreshStatisticsAfterBulk` receives actual mutations (creates + non-coalesced updates), so an all-coalesced batch no longer triggers a refresh. Gates: SQLite 4903 / PG 1998 green.

pdlug:
Follow-up fixes pushed in 4fa6c989: **true batch-local pending state** — the bulk loop now tracks the per-id running value (props the row would hold after writes queued so far) and compares each item against it, so a value-identical duplicate coalesces regardless of position (`[A,A]` on existing A → 0 writes) while last-write-wins holds (`[B,A]` → final A). Full probe matrix `[A,A]/[B,A]/[B,B]/[A,B]` pinned in the cross-backend suite for nodes (via `version`) and edges; the dirty-check now returns `{ validatedProps, unchanged }` sharing one validation with the write path. **Prettier failure** fixed (full prettier pass now clean — the earlier miss was masked by the fresh-worktree astro-types lint artifact). Gates: SQLite 4906 / PG 2001 green; graph-merge 408/408 re-verified after the shared-helper refactor.

pdlug:
Final P2 closed in de0e47ad: the edge duplicate-id probe matrix now mirrors the node table — `[A,A]/[B,A]/[B,B]/[A,B]` with exact write counts (via an `updateEdge`-counting backend wrapper that also wraps transaction-scoped backends, since edges have no version counter), final values, and per-position results, in the shared cross-backend suite. Gates: SQLite full suite green, PG 2004 green.

PR #263, opened 2026-07-14
docs: rewrite event-log guide against the post-fix transaction API (#259)

### LINKED ISSUE #259: "Materializing External Event Logs" guide: two blocking errors + gaps, from the first real implementation
Review of <https://typegraph.dev/materializing-event-logs/> from implementing it end to end in [`@nicia-ai/agent-stream-graph`](https://github.com/nicia-ai/agent-stream-graph) against 0.35.0 (the guide landed in #224 alongside the receipt work, written against requirements we identified but not yet validated by an implementation). **Every claim below was executed, not inferred.**

Headline: the individual recommendations are sound, but **two sections cannot both be followed**. Everything else is a gap rather than an error.

---

## 1. BLOCKING — "Transactional approach" is not implementable as written

> **Cursor Management** — *"**Transactional approach:** Write cursors within the same `store.transaction` callback for exactly-once materialization on supporting backends."*

You can't. On `{ history: true }` — which the guide's own Transaction Receipts and Bitemporal sections both require — `tx.sql` is a fail-loud guard (`ConfigurationError: tx.sql is not available when history capture is enabled…`), so there is no way to write a cursor row to a relational table inside `store.transaction()`.

The mechanism that **does** work is never mentioned in the guide: **`store.withRecordedTransaction(externalTx, fn)`**. Verified exactly-once (better-sqlite3, `history: true`) — a projector that throws leaves the cursor unadvanced and the node absent:

```ts
await db.run(sql`BEGIN IMMEDIATE`);            // caller owns the boundary
try {
  await store.withRecordedTransaction(db, async (tx) => {
    await tx.nodes.Item.upsertById(change.key, props);              // projection
    await db.run(sql`INSERT INTO stream_cursor … ON CONFLICT … DO UPDATE …`);  // cursor
  });
  await db.run(sql`COMMIT`);
} catch { await db.run(sql`ROLLBACK`); }        // both roll back together
```

Two things the guide must carry with that recipe:
- Write your own tables through **the external handle you passed in**, not `tx.sql` (which is unavailable here — see #252).
- On better-sqlite3 the caller frames `BEGIN`/`COMMIT` by hand; its sync driver rejects an async `db.transaction()` callback.
- `store.withTransaction()` (the non-recorded sibling) **throws** on a history store (#253).

## 2. BLOCKING — an in-graph cursor silently corrupts the receipt

The guide recommends cursors as graph nodes *and* the receipt's write count for introspecting projector outcomes. Together these are unsound: receipt counts are per-transaction with no attribution, so the cursor's own write is indistinguishable from the projector's.

A projector that **drops** a change, in a transaction that also checkpoints an in-graph cursor:

```
receipt.writes: {"nodes":{"Cursor":1},"edges":{},"total":1}
```

`writes.total > 0` no longer means "the projector wrote". The dropped change is undetectable, the cursor advances past it, and the event is lost — the precise failure the receipt exists to catch. The cursor node also lands in every `asOfRecorded` reconstruction of the belief graph. Filed as #257.

**Fix:** if the receipt is used for drop-detection, the cursor must live outside the projected graph.

## 3. The two recommendations don't compose (API gap → #255)

`transactionWithReceipt` owns its boundary (can't enlist a cursor); `withRecordedTransaction` lets you enlist a cursor but returns `T`, **not `TransactionOutcome<T>`** — no receipt, so no `writes.total` (no drop detection) and no per-transaction `recorded` (no replay anchor). Today you pick exactly-once cursors *or* receipt-anchored replay. The guide recommends both, adjacently, as though they compose. Until #255 lands, the guide should say plainly that **at-least-once + idempotent projector + receipt-anchored replay is the sanctioned path** (it's what we ship), and that the transactional path forfeits the receipt.

## 4. Never warns that `recordedNow()` is unsafe as an anchor

> *"the `recorded` anchor enables replay reconstruction at specific source offsets"*

Correct, but the guide never says **why you must take the anchor from the receipt and not from a post-commit `store.recordedNow()`**. This is the subtlest correctness trap in the pattern and our consumer had it before this upgrade: `recordedNow()` is the graph-global recorded high-water, advanced by **any** writer, so it can hand back a concurrent writer's instant and `asOfRecorded(that)` reconstructs a belief your stream never produced. We now pin it with a test that commits an interloper write in exactly that window. Worth one sentence in Transaction Receipts.

## 5. The receipt is two signals, and they deliberately disagree

The guide treats it as one ("the receipt's write count indicates completed intents"). A materializer needs both fields, and the most common real case is where they diverge:

| case | `writes.total` | `recorded` |
| --- | --- | --- |
| projector wrote | `> 0` | defined |
| delete of an absent key (legit no-op) | `1` (counts on resolve) | **`undefined`** |
| projector dropped the change | `0` | `undefined` |

A no-op delete completes a write intent but captures nothing, so its offset must carry the **prior** anchor forward — otherwise replay-by-offset breaks at exactly the offsets where nothing changed. Not derivable from the current text.

Also worth lifting from the API docs into the guide, because they bite materializers specifically: **bulk methods count by input length**, so `bulkCreate([])` counts `0` (a filter-everything-out projector must not read as a writer); and a method that **rejects** counts `0` even though, on SQLite, a failed statement does not abort the surrounding transaction — so a projector that swallows a write error and commits can persist rows the receipt never counted.

## 6. Replay never refreshes planner statistics

The guide's premise is bulk replay ("reprocessing historical events for recovery or rebuilds"), and 0.35 added auto-`ANALYZE` after large bulk writes — but the changelog is explicit that **"bulk writes inside a caller-provided transaction never auto-refresh."** A projector runs inside a transaction by construction, so the guide's headline workload is precisely the path that **never** auto-refreshes and silently keeps pre-load estimates (the changelog measures 25–200× slowdowns from exactly this). The guide should tell replay/backfill consumers to call `await store.refreshStatistics()` once after a large replay.

## 7. Smaller

- **`matchOn` is stated as though required.** *"use `getOrCreateByEndpoints` with `matchOn` policies identifying relationship fields."* The default (`matchOn: []`, endpoints-only) is the *more* idempotent choice and right for most projectors. `matchOn` **widens** the identity key, so a re-delivered edge whose matched property changed creates a **second** edge instead of converging — an idempotency footgun when misapplied.
- **"Bulk Transfer" is right but too thin to act on** — it names no function. The streaming pair is what a materializer wants, and it works: we replaced a hand-rolled O(N) query-and-re-upsert copy into a merge branch with it and got byte-identical merge results, conflicts, and provenance:
  ```ts
  await importGraphStream(branch.store, exportGraphStream(belief, { includeTemporal: true }), { onConflict: "update" });
  ```
  Two defaults are exactly right here and non-obvious, so state them: `includeDeleted` defaults to `false` (a retracted fact must not resurface in a merge), and `includeTemporal: true` is what stops the copy re-stamping every fact with the *copy's* wall clock.
- **Offsets.** The guide assumes a per-change offset. **Electric does not provide one** — every change in a catch-up batch shares `ShapeStream.lastOffset` — so a cursor may only advance at a **batch boundary**; advancing mid-batch permanently skips the batch's remaining changes on resume (`read(after)` is strictly-after). Since Electric is the motivating source, "Cursor-based resumption" should say this.

---

Happy to send a PR against the guide once #255 lands, since its outcome decides whether §1/§3 are "here's the recipe" or "here's the caveat".

pdlug:
Closes #259.

Full revision of the "Materializing External Event Logs" guide per the seven findings from its first real implementation (`@nicia-ai/agent-stream-graph`), written against the API surface of #260, #261, and #262. **Merge after those three** — the guide documents `withRecordedTransaction` returning `TransactionOutcome`, `tx.measure`, `tx.sqlAvailability`, the guard error codes, and `coalesceUnchangedUpserts`.

Per §:

1. **Cursor Management** — the unimplementable `store.transaction` cursor advice is replaced by the adopted-transaction recipe: `withRecordedTransaction`, cursor rows written through the **external handle**, examples for async drivers and better-sqlite3's manual `BEGIN IMMEDIATE` framing, and the note that `withTransaction()` on a history store is now a compile error.
2. **In-graph cursors** — the receipt-contamination trap, with both fixes: `tx.measure` scoping, or an external cursor table; plus the honest note that an in-graph cursor lands in every `asOfRecorded` reconstruction.
3. **Composition** — "The full materializer loop": adopted tx + measured projector + cursor + `receipt.recorded` anchor + drop detection. (§3 of the issue predates #255 landing; the composed loop replaces the "forfeits the receipt" caveat the issue asked for as a fallback.)
4. **`recordedNow()` warning** — anchors come from `receipt.recorded`, never a post-commit `recordedNow()` (graph-global high-water).
5. **Two-signal receipt table** — wrote / no-op delete / coalesced upsert / dropped change; carry-prior-anchor-forward rule; bulk-counts-by-input-length and rejected-counts-0 footguns.
6. **Replay statistics** — caller-provided transactions never auto-refresh; `refreshStatistics()` once after a large replay.
7. **Smaller** — `matchOn` corrected (endpoints-only default is the *more* idempotent choice; `matchOn` widens the identity key); Bulk Transfer made actionable (`importGraphStream(target, exportGraphStream(source, { includeTemporal: true }), { onConflict: "update" })`, with precise framing: `includeDeleted` defaults `false`; `includeTemporal` must be explicitly `true` — the issue's "two defaults" phrasing was slightly off); Electric batch-boundary cursor caveat (shared `lastOffset`; mid-batch advance skips the rest on resume).

Also woven in: `coalesceUnchangedUpserts` for at-least-once replay, `tx.sqlAvailability` / `isRecordedCaptureGuardError` for portable materializers, and persisted-anchor re-branding via `asRecordedInstant()`.

Every code block was verified against source on the actual PR branches (not memory); cross-link anchors verified against the sibling branches' headings. Prettier + markdownlint clean.

### DIFF
--- a/apps/docs/src/content/docs/materializing-event-logs.md
+++ b/apps/docs/src/content/docs/materializing-event-logs.md
@@ -31,10 +31,9 @@ should converge on the same graph state, not create duplicates.
 ## Idempotent Projectors
 
 Use stable source ids as TypeGraph ids whenever the source has them. For nodes,
-that usually means `upsertById`. For edges, prefer
-`getOrCreateByEndpoints` with a `matchOn` policy for the fields that identify
-the relationship. Avoid `create` in a log projector unless the source event
-itself carries a unique id you pass as the TypeGraph id.
+that usually means `upsertById`. For edges, prefer `getOrCreateByEndpoints`.
+Avoid `create` in a log projector unless the source event itself carries a
+unique id you pass as the TypeGraph id.
 
 ```typescript
 async function projectChange(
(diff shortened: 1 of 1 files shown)


pdlug:
Review fixes pushed in a14b3cea: **P1** the full materializer loop now persists the replay anchor (or carried-forward prior anchor) atomically with the cursor inside the same `dbTx` commit — `receipt.recorded` is available in-callback since `withRecordedTransaction` flushes before resolving; **P1** the bulk-copy section now states the truth: export omits tombstones and cannot round-trip `deletedAt`, so this is live-state cloning, not deletion synchronization (deletes belong in the projector); **P2** manual `refreshStatistics()` advice is scoped to custom projector loops — `importGraph`/`importGraphStream` refresh once themselves.

pdlug:
Follow-up fix pushed in 8054b96f: the guide now uses #261's final scoped-context `measure` API — `tx.measure((scopedTx) => projectChange(scopedTx, change))` — and the attribution prose is structural (a scope counts exactly the writes made through the scoped context it hands you; outer-context writes during a scope count only in the outer receipt). Swept for stale timing-based semantics; verified against the current #261 tip (9daa8bb1).

PR #268, opened 2026-07-15
feat: add Operational Identity and harden ontology

pdlug:
Adds **Operational Identity**: an opt-in, durable assertion ledger for recording when physical graph nodes refer to the same or different real-world individual.

It lets applications preserve separate source records and provenance while they:

- assert, retract, and inspect identity evidence;
- read identity classes at current, valid-time, and recorded-time coordinates;
- explicitly expand individual traversal hops through identity-class members;
- carry identity truth through transactions, interchange, schema evolution, cloning, and graph merge;
- enforce contradictions implied by explicit `different` assertions and ontology `disjointWith` relations.

Operational Identity deliberately does not perform transparent “smushing”: it does not merge properties, redirect writes, coalesce physical result rows, or silently expand ordinary queries.

## Public API

Identity is enabled and persisted at the graph level:

```typescript
const graph = defineGraph({
  id: "knowledge",
  nodes: { Person: { type: Person }, Author: { type: Author } },
  edges: {},
  identity: { sameIdAcrossKinds: "fold" },
});
```

- `"fold"` joins live nodes of different kinds that share an ID.
- `"ignore"` enables the assertion ledger without implicit cross-kind folding.
- Enabled graph types expose `store.identity`, `tx.identity`, read-only `StoreView.identity`, and identity-aware traversal options.
- Disabled graph types omit `identity` from every typed surface (store, transaction, and views), so misuse is a compile error, and run no identity locks, probes, closure work, or identity SQL at runtime (bundled backends still provision the identity tables in fresh databases). Widened or `any`-typed handles fail with `IDENTITY_NOT_ENABLED`.

```typescript
const result = await store.identity.assertSame(person, author);
// { assertion, action: "created" | "existing" }

await store.identity.membersOf(person);
await store.identity.nodesOf(person);
await store.identity.representativeOf(person);
await store.identity.areSame(person, author);
await store.identity.areDifferent(person, organization);
await store.identity.assertionsOf(person);
await store.identity.retractAssertion(result.assertion.id);
```

The write surface also includes `assertDifferent`, semantic-pair retraction, and eager bulk assertion/retraction methods. Bulk methods are eager and atomic — serialized on PostgreSQL by a per-graph advisory lock, and by SQLite's single-writer lock there. `bulkAssertSame`/`bulkAssertDifferent` return exactly one result per input pair in input order; `bulkRetractAssertions` dedupes its ids and returns only the assertions that were actually open (first-occurrence order), so its result does not align index-by-index with the input. Reasserting a current semantic pair is idempotent.

Identity expansion is opt-in per traversal hop:

```typescript
const results = await store
  .query()
  .from("Person", "person")
  .traverse("authored", "edge", { includeIdentityMembers: true })
  .to("Document", "document")
  .execute();
```

The hop considers coordinate-visible members of the source identity class, returns physical edge and node rows, preserves provenance, and deduplicates physical edge IDs within a step (a self-inverse edge between two folded peers can legitimately yield the same physical edge once per direction). Ordinary collection reads and unrelated traversal steps are unchanged.

## Semantics

The TypeGraph Identity Profile is intentionally narrower than OWL `sameAs`:

- `same` is reflexive at read time, symmetric, and transitive;
- `different` is symmetric, class-lifted, and non-transitive;
- absence of `same` evidence is not evidence of difference;
- assertions are retractable and temporal;
- self-assertions are rejected;
- `assertSame` rejects explicit or ontology-implied contradictions between the two classes;
- `assertDifferent` rejects endpoints already in the same class;
- representatives are deterministic for a read coordinate—the code-point-smallest visible `(kind, id)` wins—but are not durable entity IDs.

Current reads use a transactionally maintained closure. Historical reads reconstruct one fixed point over explicit assertions and implicit same-ID edges, then apply the requested visibility coordinate. Structurally present but coordinate-invisible bridge nodes can conduct identity without appearing in the returned member list.

Identity profile and ontology are schema-level interpretation, not a third temporal dimension. Historical views apply the Store’s pinned `sameIdAcrossKinds` mode and ontology to assertions and nodes visible at the requested coordinate. A schema migration can therefore reinterpret an older coordinate without rewriting the recorded assertion ledger.

## Lifecycle, interchange, and merge

Identity remains coherent across node creation, folding, soft/hard deletion, resurrection, bulk writes, transactions, receipts, kind removal, schema evolution, clearing, custom table names, state/archive import, working-copy cloning, and graph merge.

Creating or `upsertById()`-ing a soft-deleted `(kind, id)` resurrects the physical row, replaces its properties, and starts a new validity window. It does not revive ended identity assertions.

Interchange format 2.0 carries current identity state or archival ended assertions. Both modes bound validity windows so an imported assertion can never be snapshot-current while bypassing closure maintenance (ended rows must end in the past, open rows must start in the past). Identity import failures are recorded as `result.errors` entries with `entityType: "identity"` instead of throwing past the documented result contract. Trusted import rejects identity-enabled targets and identity-bearing input rather than silently dropping assertions or leaving closure empty; use normal `importGraphStream` for identity data.

Graph merge includes identity truth in fingerprints and diffs, deterministically deduplicates equivalent assertions, and remaps endpoints through both the canonical map and the ontology retype map. Identity conflicts are detected at plan time and surface as `GRAPH_MERGE_IDENTITY_CONFLICT` through `merge()`'s returned `Result`: direct and transitive `same`/`different` contradictions, retract/reassert races, a `different` assertion collapsed onto one reconciled survivor, an assertion naming a node another branch deletes, and assertion-id collisions — one id staged for two complete truths (validity-only divergence included) or an id the target's ledger, ended rows included, already binds to different truth. Retractions carry their complete expected row: one whose id the target reuses for different truth is skipped and reported as dropped rather than ending unrelated truth. The identity diff compares shared assertion ids by complete truth, so a hard-delete/recreate truth replacement on a branch stages as a retraction plus a new assertion and refuses typed instead of being silently invisible. A node soft-delete cascades (ending every assertion touching the node), and the deleting branch's diff stages those endings as retractions — so when the delete/modify resolution keeps the other branch's modification, retractions whose every contributing branch is explained by the overruled deletion are dropped with it (reported as `identity:deletion-overruled`), while a branch that retracted independently keeps its effect. A staged survivor's valid-time window travels with the commit write, so a branch-authored (possibly already ended) window survives resurrection instead of resetting to merge time. Merges refuse, typed, any branch whose store ran a schema operation after forking — checked against the branch's at-fork `(version, hash)` schema anchor, so even a round-trip migration that restores the document hash is refused, and revision-anchored `base@V` tokens include the active schema version to fence the same round-trip on the target side. Kind-dropping schema operations cascade the assertion ledger exactly as `removeKinds()` does — including when identity is disabled at drop time — and first enablement purges assertions naming unregistered kinds. Node resurrection and merged edges carry branch-authored valid-time windows faithfully (no inverted or reset windows), with the committed target's window winning when both sides contributed the surviving row. Convergent edits — the re-asserting branch itself also retracted the pair — merge cleanly. `MergeReport.merged` carries the identity rows the applier actually created and ended (idempotent skips excluded), and assertions the survivor rule elides are enumerated as dropped items. When a staged assertion's semantic pair is already committed on the target, the target's row always wins the survivor pick — re-derived after endpoint canonicalization, where reconciliation can collapse a branch pair onto a committed one. Merges whose only effect is identity change still advance the durable revision clock, so base@V tokens never go stale against an identity-only advance.

Commits revalidate the planned identity state inside the transaction. Incremental commits re-check live same-id peers, per-seed class/liveness fingerprints, the negative ledger, every planned assertion and retraction id, and finally re-run the full plan-time identity simulation on transaction reads, refusing plan→commit drift as a typed `BaseVersionMismatchError` replan signal; snapshot commits run the same by-id freshness check (the legacy base@V token fingerprints only current assertions, so an ended window row claiming a planned id is otherwise invisible). Any identity refusal that still escapes the commit is translated into the typed conflict error with the applier's error as its cause, never the generic merge wrapper. A property-based law suite (`tests/property/graph-merge/identity-laws.test.ts`) holds the merge to typed-failure, ledger-consistency, truth-preservation, report-coherence, and branch-effect-accounting laws over randomized identity histories — including cross-lineage chosen-id imports, hard-delete/recreate truth replacement, and same-id fold peers — in three lanes on both backends: snapshot `merge()` under each identity profile, and `mergeIncremental()` against a target that advanced after the fork. The suite has already paid for itself: it surfaced a live planner defect (a branch-minted assertion id beating the target's committed same-pair row in the survivor pick) before any reviewer did; committed target rows now always win that pick. The identity semantics of merge (plan-time derivation, contradiction simulation, commit guards) live in a dedicated `graph-merge/merge-identity.ts` module with a one-directional dependency from the merge orchestrator.

## Ontology hardening

Live and deserialized graphs now share the same validated ontology contract:

- propagate `disjointWith` through interleaved subclass and equivalence closure;
- validate source relations and recompute closures during deserialization;
- validate inverse endpoint compatibility;
- reject multiple distinct inverse partners while allowing legal self-inverses;
- reject unresolved bare extension endpoints while preserving deliberate absolute external IRIs;
- detect self-disjointness, common-subclass conflicts, equivalence/disjointness conflicts, and mixed equivalence/subclass overlaps;
- make equivalence closure iterative and size-balanced for adversarially deep graphs;
- add symmetric `getRelatedKinds` introspection;
- deprecate type-level `sameAs` and `differentFrom` in favor of individual-node Operational Identity or kind-level `equivalentTo`/`disjointWith`.

## Migration and backend guarantees

SQLite and PostgreSQL use the same query/compiler semantics and cross-backend integration suite. Identity-enabled stores require interactive atomic transactions; unsupported adapters fail at open with `IDENTITY_REQUIRES_ATOMIC_BACKEND`. Identity-disabled graphs remain supported.

PostgreSQL serializes identity-affecting writes with a per-graph advisory transaction lock. Concurrent writers to one identity-enabled graph queue behind it; other graphs and reads are unaffected.

Bundled backends provision the three identity relations—using effective custom `SqlSchema` names—before first enablement preflight. Already-enabled graphs fail closed with `IDENTITY_STORAGE_MISSING` if storage disappears; missing ledgers must be restored, while a recreated derived closure can be rebuilt before traffic resumes. PostgreSQL provisioning handles concurrent first enablement safely.

First enablement scans the graph and materializes closure. On PostgreSQL it briefly takes a shared lock on the common nodes table, so large existing graphs should be enabled during a quiet window. The very first schema commit of an identity-enabled graph is an enablement too: a legacy database populated through an unmanaged `createStore` gets the same atomic fold scan, contradiction validation, and closure build during initialization. Identity-enabled schema evolution performs a comparable rebuild.

Changing `sameIdAcrossKinds` or identity-relevant ontology is a persisted semantic migration, and a profile flip is a **breaking** schema change requiring explicit `migrateSchema()` — which runs the same closure-rebuild preflight as managed open and `evolve()`, committing rebuild and validation atomically with the schema version. While an identity change is the only breaking one pending, store construction refuses with `IDENTITY_PROFILE_MIGRATION_PENDING` rather than a generic migration error; with `autoMigrate: false`, managed open likewise refuses pending semantic changes rather than returning a Store with mismatched closure.

The public closure-rebuild operation now participates in the schema write fence, so a stale Store cannot rewrite derived identity state after another writer advances the schema.

## Performance and deferred work

Identity-disabled graphs retain a hard fast path with no identity work. Representative medians from the [benchmark report](packages/benchmarks/reports/identity-2026-07-15.md):

| Operation | SQLite | PostgreSQL |
| --- | ---: | ---: |
| Identity-disabled store construction | 0.002073 ms | — |
| Identity-disabled property update | 0.039257 ms | — |
| Folded node create | 0.137283 ms | 8.269352 ms |
| Assert/retract action | 0.103285 ms | 5.770373 ms |
| Current class read | 0.012676 ms | 0.278840 ms |

Current-coordinate class reads are O(class size), independent of total graph size: the closure lookup joins its anchor row and pins the visibility join order, so the planner cannot fall back to scanning live nodes. (PostgreSQL medians above are from a dedicated single-tenant database; earlier drafts of this table carried concurrent-lane noise.)

Caller-supplied IDs use one indexed `(graph_id, id)` folding probe; bulk creates share one batched probe, and generated IDs skip it.

Identity-expanded traversal remains correct but has two documented planner boundaries:

- current traversal reads materialized closure through indexes, while its correlated candidate-edge predicate can scan broad edge kinds per frontier row ([#270](https://github.com/nicia-ai/typegraph/issues/270));
- historical traversal embeds fixed-point reconstruction inside that predicate and has a measured quadratic reconstruction term on wide source frontiers ([#310](https://github.com/nicia-ai/typegraph/issues/310)).

Both are accepted deferred optimizations, not correctness or merge blockers. Keep broad identity-expanded traversals narrowly filtered until they land.

Other deliberate boundaries: no property substitution, canonical-row merging, write redirection, durable identity-group IDs, automatic graph-wide expansion, or semantic entity reconciliation. Zero-width assertion validity windows remain importable because the Store can legitimately emit them, and bulk receipts count intents consistently with #222.


pdlug:
Temporarily closing and reopening to retrigger CI: GitHub did not emit a synchronize workflow run for either force-update of the rebased head.

pdlug:
CI follow-up: the repeatable shard-2 worker OOM was traced to the new 20,001-member equivalence-chain regression reaching a pre-existing quadratic registry representation (one full Set copy per member, roughly 400 million entries). The reviewed fix now shares class storage and exposes self-excluding ReadonlySet views, reducing registry construction to O(N) storage while preserving iteration, forEach, Set tagging, and Node 22 set-composition semantics. A focused regression was added. Static checks and API reports pass; behavioral execution is intentionally delegated to this CI run.

PR #273, opened 2026-07-15
feat: add trusted initial graph import

pdlug:
- add `trustedImportGraph` and `trustedImportGraphStream` as an intentionally trusted initial-load path, separate from the validating interchange and collection bulk APIs
- use prepared SQLite statements and PostgreSQL `UNNEST` writes inside one transaction while temporarily deferring rebuildable secondary indexes
- switch the TypeGraph SNB benchmark loaders to the trusted path and add a reusable SQLite/PostgreSQL bulk-load investigation harness

## Contract

- requires globally empty TypeGraph node and edge tables in a fresh, dedicated database
- commits the complete stream, index rebuild, and statistics refresh atomically
- validates stream ordering and known kinds, but trusts property shapes, endpoint references, cardinality, conflicts, and IDs
- rejects history, revision tracking, uniqueness, fulltext, vectors, and adapters without the required native transactional path

## Performance

Synthetic 100,000-node / 300,000-edge initial loads:

| Backend | Existing store path | Trusted import | Speedup |
| --- | ---: | ---: | ---: |
| SQLite | 10.728s | 3.606s | 2.98x |
| PostgreSQL | 13.322s | 4.667s | 2.85x |

## Documentation

- documents the trust boundary, fresh-database requirement, atomicity, supported adapters, and first-version rejections in the interchange guide
- adds trusted initial import to the performance guidance
- includes a minor Changeset for `@nicia-ai/typegraph`


pdlug:
Benchmark follow-up (leveraging this for the SNB lanes on `bench/pggraph-comparison`):

**Write speedup confirmed on local hardware.** `pnpm bench:bulk-load` (100k nodes + 300k edges, in-memory SQLite) on an Apple-silicon machine: the public trusted path this loader uses (`trusted-public-atomic-deferred-indexes`) is **3.18× faster** than the old store `bulkInsert` path (2238ms vs 7809ms loadMs) — in line with the 2.98× synthetic figure here.

**Reads unaffected across the full query set, not just IS1–7.** The extended lanes (IC13, BFS3, IC2/IC8/IC9, GA_DEGREE, algorithm lane) all re-run `comparable=yes` at smoke through the trusted-import-loaded store. The SNB covering index is still built post-import by `materializeIndexes()` (guarded by `assertMessageIndexMaterialized`), so read behaviour is identical.

**One caveat for scale:** the SNB loader now reads the datagen CSV **twice** (a nodes pass then an edges pass, since interchange requires all nodes before all edges). The 3.18× is a pure in-memory write measurement with no CSV I/O, so it doesn't capture that second pass. Net SNB load-time win is very likely still large (write dominated at 100k), but the definitive SF1/SF10 `typegraph-sqlite` load-time A/B should run on the EC2 runner (local disk is space-constrained). If the double pass ever erodes the win below ~1.5× at SF10, spilling edges to a temp file instead of a 2nd pass is the fallback.

Observations recorded in `packages/benchmarks/reports/trusted-import-load-perf.md`.

Issue #272, opened 2026-07-15
store.algorithms: add exact weakly-connected-components (WCC)


pdlug:
Add exact weakly-connected-components to `store.algorithms`.

WCC is the concrete gap the benchmark surfaced: pgGraph computes it over the `knows` social graph in ~1ms via `graph.connected_components`, while TypeGraph had no connected-components primitive. It is exact (no float tie-breaks), so it is cross-backend deterministic and drops straight into the benchmark's byte-identical parity gate.

Implementation reuses the set-based D2 iterative substrate from #274 (label-min propagation on a node-keyed temp working table).

**Scope note:** this issue was originally an umbrella for all missing whole-graph analytics. The remaining algorithms have been split into their own issues so this one closes cleanly with the WCC implementation:
- weighted shortest path / LDBC IC14 → #275
- PageRank / Personalized PageRank → #276
- community detection / label propagation (CDLP) → #277
- local clustering coefficient (LCC) → #278

Design context for all of the above: `docs/design/sql-community-detection.md`.

PR #274, opened 2026-07-16
Fix graph traversal path explosion with set-based BFS

### LINKED ISSUE #271: store.algorithms recursive-CTE traversals enumerate paths (UNION ALL, no node dedup) — exponential blow-up with depth
## Summary

`store.algorithms` traversals (`shortestPath`, `reachable`, `neighbors`, `canReach`) compile to a recursive CTE that **enumerates paths, not a set of nodes**. Cost grows roughly geometrically with `maxHops` (~7× per hop measured) **regardless of how small the actual reachable set is**, so any traversal past ~4–5 hops on a non-sparse graph is infeasible.

## Impact

Discovered while adding an LDBC SNB IC13-style **shortest-path-between-two-persons** query to the benchmark suite (over the `knows` social graph, undirected, typical distance ~5–7 hops). TypeGraph cannot answer it at realistic depths — a single pair at `maxHops = 8` takes **44 seconds on a 31-person graph**. Every deep-traversal use of `store.algorithms` is affected; the shallow neighborhood case (depth ≤ 3) is fine.

## Evidence

`store.algorithms.reachable(seed, { edges: ["knows"], maxHops, direction: "out" })` on the committed LDBC smoke fixture (31 persons, 240 directed `knows` edges, whole component reachable by hop 3 → only **30 nodes** in the result at every depth):

| maxHops | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|
| time | 2.0 ms | 2.5 ms | 17 ms | 128 ms | 0.9 s | 6.1 s | **44.6 s** |

The result set is constant (30 nodes) from hop 3 on, yet time explodes ~7×/hop — proof the work is path enumeration, not node reachability.

## Root cause

`src/store/recursive-cte.ts`:

- `const trackPath = options.cyclePolicy === "prevent" || options.includePath;` — so the default `cyclePolicy: "prevent"` always tracks a per-row path array.
- The recursive step is `... UNION ALL ...` (no row/node de-duplication) with a per-path cycle check (`dialect.cycleCheck(n.id, r.path)`).

Net effect: one CTE row per **distinct simple path** from the source (cyclePolicy `"prevent"`), or per **walk** (cyclePolicy `"allow"`, `includePath: false` → no cycle check at all). Both are exponential in depth on a graph with average degree > 1. A social graph of average degree ~8 gives ~8^d paths at depth d.

`shortestPath` (`src/store/algorithms/shortest-path.ts`) builds the same CTE with `includePath: true` and its cost comment ("under cyclePolicy `prevent`, |reachable| ≤ V, so this is bounded by V × p") describes the *final de-duplicated* result, not the CTE it materializes — the CTE itself is the number of simple paths, not V. That comment (and the "minimum discovered depth" framing in `reachable`'s docs) is misleading about the actual cost.

## Suggested direction

A true BFS/Dijkstra reachability needs **per-node** de-duplication keeping the minimum depth, which a plain `UNION ALL` recursive CTE can't express (it can only dedupe whole rows, and rows carry depth/path). Options:

1. Set-based fixpoint that dedupes by node id keeping `min(depth)` between iterations (iterative frontier expansion in the driver, or a CTE shaped so `UNION` dedupes on node id alone).
2. For `shortestPath` specifically, a bidirectional or early-terminating BFS that stops when the target is first dequeued.
3. At minimum, document the current algorithm as path-enumeration with an explicit depth/degree ceiling, and have deep traversals fail fast rather than hang.

## Repro

Load the smoke fixture into a SQLite store and time `store.algorithms.reachable(seed, { edges: ["knows"], maxHops, direction: "out" })` for `maxHops` 2…8 (see table above). `neighbors`/`shortestPath`/`canReach` share the same CTE and the same blow-up.

## Downstream

Blocks the benchmark comparison's IC13 (shortest-path) lane on the TypeGraph side (`bench/pggraph-comparison`). The BFS3 (depth-3 neighborhood) lane is unaffected and runs fine. Once the algorithm does per-node dedup, that branch rebases and enables IC13 on TypeGraph with no driver changes.

pdlug:
- replace `reachable`, `neighbors`, `shortestPath`, and `canReach` recursive path enumeration with set-based BFS
- add the iterative graph-operation substrate: a host round loop over a node-keyed temporary working table with convergence, snapshot, bind-budget, and cleanup ownership
- use bidirectional BFS for shortest path and reachability checks, preserving `(kind, id)` identity and temporal semantics
- retain a chunked inline frontier for bounded traversal on backends without a pinned transactional connection
- add an internal branded temporary-statement seam so history-enabled stores can use ephemeral working state without reopening arbitrary raw writes

## Why

The recursive CTEs admitted one row per simple path and only grouped after recursion. Dense cyclic graphs therefore produced combinatorial intermediate work even though the public algorithms return node-set or shortest-path results.

This change admits each `(kind, id)` once at its minimum depth. Work is now bounded by reached nodes and incident edges rather than the number of paths. It also establishes the iterative execution shape described in the SQL community-detection design, so WCC/CDLP/PageRank can reuse the same working-table lifecycle instead of introducing a second iteration engine.

## Backend behavior

- SQLite uses a deferred snapshot for semantically read-only temporary work and leaves the main database writer slot free.
- PostgreSQL/PGlite use repeatable read on one pinned connection; an unforgeable internal capability permits temporary DDL/DML while permanent graph/history state remains protected.
- Temporary tables are dropped in `finally`; rollback owns cleanup after an aborted PostgreSQL statement, without masking the original error.
- Non-transactional backends use bind-limit-aware `VALUES` frontiers for these bounded algorithms.

## Performance

Reproduced issue #271 with its committed LDBC smoke fixture: 31 people, 240 directed `knows` edges, source person `0`, and `reachable(..., maxHops: 2..8)` through the public API. Result cardinality was unchanged (28 nodes at depth 2; all 30 component nodes from depth 3 onward).

Measured on an Apple M4 Pro with Node 26.5.0 and in-memory SQLite. The same-machine pre-fix run used commit `92479d44` and one sweep, matching the original issue's methodology; the PR column is the median of 20 sweeps at `4775532b` with p95 in parentheses.

| maxHops | Original issue | Same-machine pre-fix | This PR median (p95) | Same-machine speedup |
|---:|---:|---:|---:|---:|
| 2 | 2.0 ms | 1.007 ms | 0.499 ms (0.765) | 2x |
| 3 | 2.5 ms | 1.518 ms | 0.772 ms (1.028) | 2x |
| 4 | 17 ms | 10.061 ms | 0.940 ms (1.176) | 11x |
| 5 | 128 ms | 70.276 ms | 0.950 ms (1.237) | 74x |
| 6 | 0.9 s | 466.813 ms | 0.952 ms (1.074) | 490x |
| 7 | 6.1 s | 3.282 s | 0.942 ms (1.176) | 3,484x |
| 8 | **44.6 s** | **20.311 s** | **0.938 ms (1.020)** | **21,652x** |

The absolute pre-fix time is lower on this machine than in the original report, but it reproduces the same failure shape: approximately 6–7x more work per added hop from depth 3 onward. The PR is flat after the component converges. At depth 8 it is 47,543x faster than the original issue measurement and 21,652x faster than the same-machine pre-fix run.

## Validation

- `pnpm fix`
- `pnpm typecheck`
- `pnpm test` — 247 files passed, 4,978 tests passed
- `pnpm test:postgres` — 71 files passed, 2,039 tests passed
- dense cyclic graph, cross-kind same-ID identity, bind-budget, bounded-work, history-enabled, failed-initialization cleanup, SQLite writer-slot, PostgreSQL access-mode, and PGlite temp-leak coverage
- exact issue #271 LDBC fixture benchmark described above

Closes #271

### DIFF
--- a/.changeset/set-based-graph-traversal.md
+++ b/.changeset/set-based-graph-traversal.md
@@ -1,0 +1,13 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Replace path-enumerating recursive CTEs in `reachable`, `neighbors`,
+`shortestPath`, and `canReach` with set-based breadth-first search.
+
+Transactional SQLite and PostgreSQL backends now execute graph iterations
+against a connection-local temporary working table, de-duplicated by node kind
+and ID on every round. Non-transactional backends retain parity through a
+bind-limit-aware inline frontier. Traversals run in one snapshot where the
+backend supports transactions, preserve temporal filtering, and clean up
+temporary state on success or failure.
--- a/apps/docs/src/content/docs/examples/document-management.md
+++ b/apps/docs/src/content/docs/examples/document-management.md
@@ -684,9 +684,8 @@ async function getBreadcrumb(
 }
 ```
 
-`reachable` returns `{ id, kind, depth }` — one recursive-CTE query returns
-the full ancestor chain, then a single batched `getByIds` hydrates the folder
-properties.
+`reachable` returns `{ id, kind, depth }` from a set-based BFS frontier, then a
+single batched `getByIds` hydrates the folder properties.
 
 ## Bulk Operations
 
--- a/apps/docs/src/content/docs/graph-algorithms.md
+++ b/apps/docs/src/content/docs/graph-algorithms.md
@@ -16,9 +16,13 @@ store.algorithms.neighbors(alice, { edges: ["knows"], depth: 2 });
 store.algorithms.degree(alice, { edges: ["knows"] });
 ```
 
-Each call compiles to a single `WITH RECURSIVE` CTE (or a plain `COUNT` for
-`degree`), using the same path-tracking and cycle-detection machinery as
-`.recursive()` and `store.subgraph()`. The algorithms work identically on
+Traversal calls use a set-based breadth-first frontier. Each level expands the
+current `(id, kind)` set with a bind-limit-aware SQL query. Transactional
+backends keep the visited set in a temporary working table; backends without a
+pinned transaction use a chunked inline working relation. Each node is admitted
+only at its minimum depth. `shortestPath` and `canReach` search from both
+endpoints and stop when the frontiers meet.
+`degree` remains a single `COUNT` query. The algorithms work identically on
 SQLite and PostgreSQL.
 
 ## When to Reach for Algorithms
@@ -46,13 +50,14 @@ Every traversal algorithm takes the same base options:
 | `edges` | `readonly EdgeKinds<G>[]` | *(required)* | Edge kinds to follow |
 | `maxHops` | `number` | `10` | Maximum traversal depth (1 – 1000) |
 | `direction` | `"out" \| "in" \| "both"` | `"out"` | Edge direction |
-| `cyclePolicy` | `"prevent" \| "allow"` | `"prevent"` | Skip visited nodes per path |
+| `cyclePolicy` | `"prevent" \| "allow"` | `"prevent"` | Compatibility option; both values use set-based node de-duplication |
 | `temporalMode` | `TemporalMode` | `graph.defaults.temporalMode` | Filter applied to nodes and edges along the traversal — see [Temporal Behavior](#temporal-behavior) |
 | `asOf` | `string` (ISO-8601) | *(none)* | Snapshot timestamp, required when `temporalMode: "asOf"` |
 
-`direction: "both"` treats edges as undirected. `cyclePolicy: "allow"` skips
-cycle tracking and relies solely on `maxHops` to terminate — use it only
-when you know the graph is acyclic or you want maximum raw performance.
+`direction: "both"` treats edges as undirected. `cyclePolicy` remains accepted
+for compatibility with recursive query-builder traversals, but it does not
+change these algorithms: their node-set results and shortest paths never need
+to revisit a node.
 
 ## shortestPath
 
@@ -102,8 +107,8 @@ sorted by ascending depth.
 
 ## canReach
 
-Fast boolean check that short-circuits with `LIMIT 1` so the database stops
-traversing as soon as it finds the target.
+Fast boolean check that uses the same balanced bidirectional BFS as
+`shortestPath` and stops as soon as the two visited sets meet.
 
 ```typescript
 const connected = await store.algorithms.canReach(alice, bob, {
@@ -112,8 +117,8 @@ const connected = await store.algorithms.canReach(alice, bob, {
 });
 ```
 
-Use this when you only care whether a path exists — it's cheaper than
-`shortestPath` because it never decodes the path.
+Use this when you only care whether a path exists. It shares the same search
+cost as `shortestPath` but does not return the reconstructed path.
 
 ## neighbors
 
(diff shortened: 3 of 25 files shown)


pdlug:
## Closes #271 correctly
- **Set-based min-depth admission** is right: `ON CONFLICT (graph_id, run_id, side, node_kind, node_id) DO NOTHING` (+ `NOT EXISTS`) admits each `(kind,id)` once at its first/minimum depth — work is bounded by reached nodes/incident edges, not simple paths. The flat ~0.94ms across depths 2–8 on the exact fixture (vs 44.6s) is the shape I'd expect.
- **Bidirectional shortest path is correct**, including the subtle part: `findWorkingMeeting` selects the **minimum** combined depth (`ORDER BY total_depth … LIMIT 1`, deterministic tie-break) over all accumulated nodes each round, and the reverse side expands with `reverseDirection(direction)` — so directed shortest-path semantics hold.
- **Public signatures unchanged** (shortestPath/reachable/neighbors/canReach, ReachableNode, ShortestPathResult); `cyclePolicy` correctly becomes a no-op.

## Benchmark impact 
This un-gates three queries on TypeGraph with no code changes — just removing them from `TYPEGRAPH_UNSUPPORTED`: **IC13** (shortestPath), **GA_BFS** and **GA_SSSP** (reachable at whole-component `maxHops=64` now converges when the frontier empties, ~7 rounds on the SF1 knows component). Once merged I'll un-gate + run smoke parity — my IC13/GA_BFS/GA_SSSP digests are checked againstneo4j/ladybug, so that's an **independent cross-engine correctness gate** on this rewrite for free. Only GA_WCC stays gated (needs #272) — and this PR's substrate is exactly what unblocks it.

## Community-detection connection — this is the shared foundation, done right
`iterative-graph-operation.ts` realizes the **primitive** (host round loop over a node-keyed temp working table, convergence + snapshot + cleanup ownership). So WCC/CDLP/PageRank reuse this lifecycle rather than a second iteration engine — precisely the 'build once, use for both #271 and #272' path. 

## Non-blocking notes
1. **The temp-statement seam is trust-by-convention.** `asCompiledTemporaryStatementSql` is a pure type cast and `executeTemporaryStatement` runs the SQL directly; 'cannot mutate graph/history tables' is guaranteed by encapsulation (the `INTERNAL_TEMPORARY_WRITES` Symbol + the brand only ever applied to temp SQL), not runtime inspection — same trust model as #273's trusted import, which is fine, but history integrity now depends on that discipline holding forever. Suggest the recorded-capture-capability-gate test explicitly assert a forged capability / non-temp statement can't reach permanent tables.
2. **Per-call temp-table create/drop** adds a fixed overhead to every algorithm invocation — correct and honestly measured, just worth noting for high-QPS algorithm use vs the old single-CTE call.
3. **Bidirectional termination** is the one place to keep a targeted test: an asymmetric-frontier case with multiple meeting nodes at different combined depths (odd vs even path length) to lock in that first-meeting = global-min. The dense-cyclic test likely covers it; just confirm.

Net: unblocks the benchmark's traversal + BFS/SSSP lanes on TypeGraph and lays the groundwork for the algorithm gaps in #272.

PR #282, opened 2026-07-16
perf(query): index nodes by (graph_id, id) for bare-id lookups (#280)

### LINKED ISSUE #280: store.algorithms.degree scans all nodes on SQLite (nodeKindSubquery can't seek the kind-leading PK)
## Symptom

At LDBC SF1, `store.algorithms.degree(personId, { edges: ['knows'], direction: 'out' })` takes **~95ms on SQLite** (vs 0.2ms on PostgreSQL, 0.3ms ladybug, 0.1ms pgGraph). Stable p50/p95, not noise.

## Root cause (EXPLAIN-confirmed on a real SF1 store)

```
SEARCH typegraph_edges USING COVERING INDEX typegraph_edges_from_idx (graph_id=? AND from_kind=? AND from_id=? AND kind=?)   -- edge lookup: fine
SCALAR SUBQUERY 1
  SEARCH typegraph_nodes USING COVERING INDEX sqlite_autoindex_typegraph_nodes_1 (graph_id=?)                               -- seeks graph_id ONLY
```

`compileDirectionFilter` resolves the seed's kind via `nodeKindSubquery`: `SELECT kind FROM typegraph_nodes WHERE graph_id=? AND id=? LIMIT 1`. The nodes PK is **`(graph_id, kind, id)`** (schema/sqlite.ts), and every other node index leads with `kind`/`deleted_at`/`valid_from` — so a bare `graph_id + id` lookup has no seekable index. SQLite seeks `graph_id` (the whole graph, ~3M nodes at SF1) and scans them filtering `id`. The literal-`from_kind` variant (no subquery) is a pure index seek in sub-ms, confirming the subquery scan is the entire cost.

The subquery was added (correctly) to make the *edge* index seekable via `from_kind` — but on SQLite it trades an edge scan for a nodes scan, because you can't seek a node by id when the id-carrying index leads with kind.

## Impact

Degree-specific (reachable/shortestPath/neighbors resolve kinds from the working table / edge rows, not this subquery — BFS is 425ms at SF1, not scan-dominated). Any future algorithm using `nodeKindSubquery` inherits it.

## Fix options

1. **Add a `(graph_id, id)` secondary index on `typegraph_nodes`** — serves any bare-id node lookup; the subquery then seeks in sub-ms. Small; PostgreSQL is already fast so this is effectively SQLite-targeted (but harmless on PG). Cost: one more index maintained per node write.
2. Resolve the kind in a separate point read via the query-builder's fast path before compiling the degree query (extra round trip, no schema change).

Found via the SF1 pgGraph-comparison benchmark (bench/pggraph-comparison).

pdlug:
Closes #280.

## Problem

The nodes primary key is `(graph_id, kind, id)`. A lookup that knows a node's `id` but **not** its `kind` cannot seek that key. `store.algorithms.degree`'s node-kind subquery (`SELECT kind FROM typegraph_nodes WHERE graph_id = ? AND id = ? LIMIT 1`) does exactly that, so SQLite falls back to a `graph_id`-only scan of every node in the graph — ~3.16M nodes at LDBC SNB SF1 — making `degree()` ~95 ms per call.

The same probe exists at a **recorded coordinate**: `store.asOfRecorded(rt).degree(...)` swaps the node source to `typegraph_recorded_nodes`, whose entity index also leads with `kind` — and the recorded partition is strictly larger (one row per historical version), so the scan is worse there.

## Fix

Add a `(graph_id, id)` index to **both node relations** in **both dialect schemas**:

- `typegraph_nodes_id_idx` on the live nodes table
- `typegraph_recorded_nodes_id_idx` on the recorded nodes table (parity for recorded-pinned reads)

## Verification (real SF1, 3.16M nodes)

| | before | after |
| --- | --: | --: |
| bare-id lookup plan | scan of 3.16M nodes (`graph_id=?` only) | `SEARCH … USING INDEX typegraph_nodes_id_idx (graph_id=? AND id=?)` |
| `store.algorithms.degree` | ~95 ms | **0.06–0.49 ms** (~200–1500×) |

Validated at SF1 in the full suite re-run: GA_DEGREE 102 ms → **0.05 ms** on tg-sqlite. PostgreSQL already coped via its planner (~0.2 ms) but is indexed too, so the bare-id access path is explicit rather than planner-luck-dependent — and the PG EXPLAIN for IC9 confirmed `typegraph_nodes_id_idx` is used for the bare-id join there as well.

## Existing databases (upgrade path)

This index does **not** reach an already-initialized database automatically: `bootstrapTables()` only runs on first boot (missing-table), and `createStore`/`createVerifiedStore` are zero-DDL by contract. To adopt the indexes on an existing deployment, run once:

```ts
await backend.bootstrapTables();
```

Every bootstrap statement is `CREATE … IF NOT EXISTS`, so the call is idempotent and only the two new indexes are created. Caveats: on PostgreSQL this issues a plain `CREATE INDEX` (briefly locks writes on large tables — run in a maintenance window, or apply the equivalent `CREATE INDEX CONCURRENTLY` statements manually), and it should not be raced from many replicas simultaneously (concurrent multi-table DDL can deadlock on Postgres SHARE locks). A declarative system-index materialization path (riding the existing `typegraph_index_materializations` machinery) is planned as a follow-up so future base-index additions roll out without this manual step.

## Tests

- DDL-presence assertions for both new indexes in both backend suites.
- **Query-plan assertions** (`EXPLAIN QUERY PLAN`): `degree()` now must seek `typegraph_nodes_id_idx` (no `SCAN typegraph_nodes`) at the live coordinate, and `typegraph_recorded_nodes_id_idx` (no `SCAN typegraph_recorded_nodes`) at a recorded pin, in `tests/degree-index-alignment.test.ts`.
- **Upgrade-path tests** (SQLite + PostgreSQL): start from the pre-PR physical schema (indexes dropped), run `backend.bootstrapTables()` once, assert both indexes exist.
- Full SQLite suite green; PostgreSQL lane green. `pnpm fix` + `pnpm typecheck` clean.

### DIFF
--- a/.changeset/bare-id-node-index.md
+++ b/.changeset/bare-id-node-index.md
@@ -1,0 +1,17 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Add a `(graph_id, id)` index to the live and recorded node tables so bare-id
+lookups (a node's `id` without its `kind`) seek instead of scanning the
+graph's node partition — the composite keys lead with `kind`, so they can't
+serve that probe. `store.algorithms.degree()`'s node-kind subquery is the
+main consumer: ~95 ms → sub-millisecond at LDBC SNB SF1 (3.16M nodes) on
+SQLite, at the live and recorded coordinates alike.
+
+New databases get both indexes at bootstrap. Existing databases adopt them
+with a one-time `await backend.bootstrapTables()` — every statement is
+`CREATE … IF NOT EXISTS`, so the call is idempotent and only creates what's
+missing. On PostgreSQL this issues a plain `CREATE INDEX` (briefly locks
+writes on large tables); schedule it, or apply the equivalent
+`CREATE INDEX CONCURRENTLY` statements manually.
--- a/packages/typegraph/src/backend/drizzle/schema/postgres.ts
+++ b/packages/typegraph/src/backend/drizzle/schema/postgres.ts
@@ -125,6 +125,11 @@ export function createPostgresTables(
       ),
       index(`${n.nodes}_deleted_idx`).on(t.graphId, t.deletedAt),
       index(`${n.nodes}_valid_idx`).on(t.graphId, t.validFrom, t.validTo),
+      // Bare-id lookup: the primary key leads with `kind` (graph_id, kind, id),
+      // so a `WHERE graph_id = ? AND id = ?` probe that doesn't know the kind
+      // can't seek it. `store.algorithms.degree`'s node-kind subquery resolves a
+      // seed's kind by id, so it depends on this access path. See typegraph#280.
+      index(`${n.nodes}_id_idx`).on(t.graphId, t.id),
       ...buildPostgresNodeIndexBuilders(t, indexes),
     ],
   );
--- a/packages/typegraph/src/backend/drizzle/schema/postgres.ts
+++ b/packages/typegraph/src/backend/drizzle/schema/postgres.ts
@@ -242,6 +242,12 @@ export function createPostgresTables(
         t.validFrom,
         t.validTo,
       ),
+      // Bare-id lookup parity with the live nodes table: recorded-pinned
+      // reads swap this relation in as the node source, and `_entity_idx`
+      // leads with `kind`, so the same kind-by-bare-id probe (e.g.
+      // `degree()` at a recorded coordinate) would otherwise scan every
+      // historical version in the graph. See typegraph#280.
+      index(`${n.recordedNodes}_id_idx`).on(t.graphId, t.id),
     ],
   );
 
--- a/packages/typegraph/src/backend/drizzle/schema/sqlite.ts
+++ b/packages/typegraph/src/backend/drizzle/schema/sqlite.ts
@@ -123,6 +123,14 @@ export function createSqliteTables(
       ),
       index(`${n.nodes}_deleted_idx`).on(t.graphId, t.deletedAt),
       index(`${n.nodes}_valid_idx`).on(t.graphId, t.validFrom, t.validTo),
+      // Bare-id lookup: the primary key leads with `kind` (graph_id, kind, id),
+      // so a `WHERE graph_id = ? AND id = ?` probe that doesn't know the kind
+      // can't seek it — SQLite falls back to a graph_id-only scan of every node
+      // in the graph (~3M at LDBC SF1). `store.algorithms.degree`'s
+      // node-kind subquery does exactly that lookup (it resolves the seed's
+      // kind by id), so without this index degree is a full nodes scan (~95ms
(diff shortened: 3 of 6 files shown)


pdlug:
## Review + fixes applied (commit 0add0b14)

Two findings from review, both confirmed against the code and fixed directly on this branch. The PR description has been rewritten to match the current state (it feeds the squash merge).

### P1 (confirmed): the "picked up idempotently on next open" claim was false

`bootstrapTables()` only runs when reading the schema throws a missing-table error (`src/schema/manager.ts:121-134`) — i.e. first boot on an empty database. `createStore`/`createVerifiedStore` are zero-DDL by contract, and `schema/migration.ts` never emits index actions, so an already-initialized database would **never** receive this index and would keep the 3.16M-row scan. (The SF1 benchmark didn't notice because it imports fresh databases.)

**Fix:** rather than bolting on a one-off ensure path, the description now documents the honest upgrade story — a one-time explicit `await backend.bootstrapTables()` (all statements are `CREATE … IF NOT EXISTS`, so it's idempotent and only creates what's missing), with the PG locking + multi-replica caveats spelled out. New upgrade-path tests on both SQLite and PostgreSQL start from the pre-PR physical schema (indexes dropped), run the adoption call, and assert both indexes exist. A declarative system-index materialization path (riding the existing `typegraph_index_materializations` machinery, so future base-index additions roll out without a manual step) is agreed as a follow-up — this PR intentionally does not start that abstraction piecemeal.

### P2 (confirmed): recorded-time `degree()` kept the scan

Recorded-pinned reads swap `typegraph_recorded_nodes` in as the node source (`recordedReadSchemaFor`, `src/query/compiler/schema.ts:434`), and every recorded-node index leads with `kind` or `recorded_to` — so `store.asOfRecorded(rt).degree(...)`'s kind-by-bare-id probe scanned the recorded partition, which is strictly larger than the live one (a row per historical version).

**Fix:** added `typegraph_recorded_nodes_id_idx (graph_id, id)` in both dialect schemas, plus a recorded-coordinate plan test.

### Test hardening beyond the findings

The original tests only asserted the index appears in generated DDL — they'd pass even if the planner never used it. `tests/degree-index-alignment.test.ts` now asserts the actual `EXPLAIN QUERY PLAN`: `degree()` must seek `typegraph_nodes_id_idx` (no `SCAN typegraph_nodes`) at the live coordinate and `typegraph_recorded_nodes_id_idx` (no `SCAN typegraph_recorded_nodes`) at a recorded pin.

### Verification

- Full SQLite suite: 4,997 passed.
- PostgreSQL lane: 71 files / 2,046 tests passed (includes the new PG upgrade-path test).
- `pnpm fix` + `pnpm typecheck` clean.

PR #283, opened 2026-07-16
perf(query): late materialization for ORDER BY … LIMIT queries (#281)

### LINKED ISSUE #281: Query compiler: no late materialization — projected columns fetched for all rows before ORDER BY … LIMIT
## Summary

For an ordered, limited query (`.orderBy(...).limit(k)`), the query compiler emits a single flat `SELECT <all projected columns> FROM ...joins... WHERE ... ORDER BY <keys> LIMIT k`. Every projected column — including large ones not referenced by the sort keys, join keys, or predicates — is fetched and materialized for **every candidate row**, carried through the sorter, and then all but `k` rows are discarded.

The compiler has no *late materialization* (a.k.a. deferred projection / "top-K then fetch"): sort + limit on the minimal column set first, then re-join to fetch the wide columns for only the `k` survivors.

## Evidence

Measured on the LDBC SNB SF1 dataset with an IC9-style query — the 20 most recent `Comment`s (`ORDER BY creationDate DESC, id DESC LIMIT 20`) authored by a person's 2-hop `knows` neighborhood:

- Candidate set (before `LIMIT`): **1,184,787 comments** (3,662 author persons × ~324 comments each).
- Same query, `content` in the projection vs. sort-keys-only:

  | projection | latency |
  | --- | --: |
  | `{ id, content, creationDate, creator... }` (full) | **2,928 ms** |
  | `{ id, creationDate }` (sort keys only) | **1,958 ms** |

  → fetching `content` for all 1.18M candidates costs **~970 ms (~33%)** even though only 20 rows survive `LIMIT 20`. Pure waste.

`EXPLAIN QUERY PLAN` on SQLite shows the wide projection carried into `USE TEMP B-TREE FOR ORDER BY`, plus a per-candidate heap fetch for the non-covered `content` column. PostgreSQL is expected to pay this far more heavily (wide-row heap fetches + a sort that spills to disk when the projected rows don't fit `work_mem`).

## Proposed fix

Late materialization when a plan has both a `Sort` and a `LimitOffset` and the projection references columns beyond those needed for the sort keys / join keys / predicates:

1. Inner query: `SELECT <sort keys + node identity (graph_id, kind, id)> FROM ...joins... WHERE ... ORDER BY <keys> LIMIT k`.
2. Outer query: join the inner result back to the node table(s) on identity to fetch the full projection for only the `k` surviving rows.

Backend-neutral (standard-SQL subquery/CTE), so it stays on the single shared compile path with no per-dialect branching. Skip the transform when the projection is already minimal (no non-key columns) or entirely covered by an index — no benefit there.

## Scope / notes

- Motivated by IC9 but general: any `ORDER BY … LIMIT` over a large candidate set with a wide projection benefits (feeds, recent-activity, top-N-by-score).
- The residual ~1,958 ms (sort-keys-only) is a separate, architectural cost — the top-K-recent-over-a-dynamic-author-set pattern has no supporting `(creator, creationDate)` index in the generic edge schema; that is out of scope for this issue.
- Correctness must hold across SQLite + PostgreSQL, temporal-validity filters, soft-delete, and multi-hop traversal joins; covered by the shared cross-backend suite.

pdlug:
Closes #281.

## Problem

A selective query with `ORDER BY … LIMIT` compiled to a single flat `SELECT` that projected **every** column — including large ones not needed for ordering, such as a `content` prop — for **every candidate row**, carried them through the sorter, then discarded all but the `LIMIT` survivors.

At LDBC SNB SF1, IC9's Comment leg fans out to **1,184,787 candidate comments** for a top-20; `content` was extracted for all 1.18M and sorted, even though only 20 survive.

## Fix

Add a late-materialization (deferred-projection) branch to the standard compile path, alongside the existing count-aggregate fast path:

1. Build a **lean** candidate CTE set carrying only identity + ordering + predicate columns (the projection-only columns are dropped via `collectRequiredColumnsByAlias(..., { includeProjection: false })` + a pruned `selectiveFields`).
2. A `cte_lm_topk` CTE sorts + limits that lean set, carrying each sort key out as `__lm_sk{n}`.
3. The outer `SELECT` re-joins the physical `nodes` table by `(graph_id, kind, id)` for each projected alias and fetches the deferred columns **for only the surviving rows**, re-ordering by the carried sort keys.

The logical plan is reconciled with `stripLimitOffsetFromPlan` (the LIMIT moves into the inner CTE), exactly as the count fast path does. The start + traversal candidate-CTE prefix is now built by a helper (`buildStandardStartAndTraversalCtes`) shared with the flat plan rather than a third inline copy.

### Properties

- **Backend-neutral** — a plain `SELECT … FROM (topK) JOIN nodes ON identity` is standard SQL on both dialects; no new `DialectAdapter` member, no `=== "sqlite"` branching under `src/query`.
- **Live-table only** — the outer identity re-join is sound because `(graph_id, kind, id)` is the live nodes PK. Recorded-time reads (`ast.recordedAsOf`) swap the node source to the recorded history relation, where that key matches **one row per version** — the re-join would duplicate survivors and read stale-version props. Recorded reads therefore fall back to the flat plan, guarded by a regression test that fails with duplicated rows if the gate is removed.
- **Collapse-aware** — handles the common rowset-collapsed selective-traversal shape (all identities carried into one CTE) as well as the separate-CTE and start-only shapes.
- **Tightly gated** — fires only for the selective `.select()` path with `ORDER BY` + a positive `LIMIT` and at least one deferrable column, at the live coordinate. Aggregates, vector/fulltext, optional (LEFT JOIN) traversals, edge-field projections, non-selective queries, and recorded-time reads fall back to the flat plan unchanged.

### Ordering semantics

The outer `SELECT` re-orders survivors by the carried `__lm_sk{n}` sort keys with the same `(col IS NULL)` null-placement pattern as the flat plan. When the `ORDER BY` determines a unique top-K, results are identical to the flat plan. When ties span the LIMIT boundary, both plan shapes pick an arbitrary tied subset (SQL non-determinism the flat plan already has) and may pick differently.

## Results (LDBC SNB SF1, real data)

IC9 Comment leg (top-20 over the friend/FoF message fan-out), `late-mat` confirmed active on the executed SQL:

| seed (FoF authors) | before | after | speedup |
| --- | --: | --: | --: |
| 2,089 | 1,674 ms | **1,180 ms** | ~30% |
| 3,374 | 2,737 ms | **1,893 ms** | ~31% |
| 3,662 | 2,968 ms | **2,108 ms** | ~29% |

`content` is now extracted for 20 rows instead of ~1.18M — matching the ~33% content-materialization cost measured for #281. The residual cost is the inherent candidate fan-out (a separate, architectural item noted in the issue). Validated at SF1 in the full suite re-run: IC9 4,325 ms → 2,718 ms (~37%) on tg-sqlite.

## Tests

- New shared cross-backend suite `tests/backends/integration/late-materialization.ts` — 9 cases (top-K, OFFSET, predicate on the deferred node, null ordering, tie-breaks, start-only/non-collapse path, multi-hop, limit-exceeds-set, **recorded-time fallback with multi-version history**) run on SQLite, libSQL, and PGlite.
- The recorded-time case was verified to fail (4 rows returned for `limit(2)`, duplicated survivors) when the recorded gate is disabled.
- Full SQLite suite green (**5,022 passed**); PostgreSQL lane green (**2,072 passed**); existing `orderBy + limit + select` tests now route through the new branch with no changes.
- `pnpm fix` + `pnpm typecheck` clean.

### DIFF
--- a/.changeset/late-materialization-order-by-limit.md
+++ b/.changeset/late-materialization-order-by-limit.md
@@ -1,0 +1,16 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Selective `ORDER BY … LIMIT` queries now compile with late materialization:
+the query sorts and limits a lean candidate set carrying only identity, sort
+keys, and predicate columns, then re-fetches the deferred projection columns
+by primary key for only the surviving rows — instead of extracting every
+projected column for every candidate and discarding all but the `LIMIT`
+survivors after the sort. At LDBC SNB SF1, IC9's top-20 over a 1.18M-comment
+fan-out stops extracting `content` 1.18M times, ~30–37% faster on SQLite.
+
+The transform fires only on the selective `.select()` path with `ORDER BY`
+and a positive `LIMIT` at the live coordinate. Aggregates, vector/fulltext,
+optional (LEFT JOIN) traversals, edge-field projections, non-selective
+queries, and recorded-time reads keep the flat plan unchanged.
--- a/packages/typegraph/src/query/compiler/emitter/index.ts
+++ b/packages/typegraph/src/query/compiler/emitter/index.ts
@@ -18,6 +18,9 @@ export {
   type StandardQueryEmitterInput,
 } from "./standard";
 export {
+  buildLateMaterializedOuterOrderBy,
+  buildLateMaterializedOuterProjection,
+  buildLateMaterializedTopKCte,
   buildLimitOffsetClause,
   buildStandardEmbeddingsCte,
   buildStandardFromClause,
@@ -32,5 +35,8 @@ export {
   buildStandardStartCte,
   buildStandardTraversalCte,
   buildStandardVectorOrderBy,
+  LATE_MAT_TOPK_CTE_ALIAS,
+  lateMaterializedPhysicalAlias,
+  lateMaterializedProjectedNodeAliases,
   type StandardEmitterPredicateIndex,
 } from "./standard-builders";
--- a/packages/typegraph/src/query/compiler/emitter/standard-builders.ts
+++ b/packages/typegraph/src/query/compiler/emitter/standard-builders.ts
@@ -810,6 +810,171 @@ export function buildStandardOrderBy(
   return sql`ORDER BY ${sql.join(parts, sql`, `)}`;
 }
 
+// ============================================================
+// Late materialization (deferred projection)
+// ============================================================
+//
+// For an ORDER BY … LIMIT query whose projection reads columns beyond the
+// ordering/identity keys, the flat plan carries every projected column (e.g. a
+// large `content` prop) through the sorter for every candidate row and then
+// discards all but the LIMIT survivors. Late materialization sorts+limits a
+// *lean* candidate set (identity + sort keys only), then re-joins the physical
+// node table by `(graph_id, kind, id)` to fetch the deferred columns for only
+// the surviving rows. See the standard emitter's late-materialization branch
+// for the eligibility gate; these builders assume it already held.
+
+export const LATE_MAT_TOPK_CTE_ALIAS = "cte_lm_topk";
+const LATE_MAT_SORT_KEY_PREFIX = "__lm_sk";
+const LATE_MAT_PHYSICAL_ALIAS_PREFIX = "lm_";
+
+export function lateMaterializedPhysicalAlias(alias: string): string {
+  return `${LATE_MAT_PHYSICAL_ALIAS_PREFIX}${alias}`;
+}
+
+/** Node aliases in the query: the start alias plus each traversal's node. */
+export function lateMaterializedNodeAliases(ast: QueryAst): readonly string[] {
+  return [
+    ast.start.alias,
+    ...ast.traversals.map((traversal) => traversal.nodeAlias),
+  ];
+}
+
+/**
+ * Node aliases the projection actually reads — the physical tables the outer
+ * SELECT must re-join. Edge aliases are gated out upstream, so every projected
+ * alias resolves to a node table here.
+ */
+export function lateMaterializedProjectedNodeAliases(
+  ast: QueryAst,
+): readonly string[] {
+  const referenced = new Set<string>();
+  for (const field of ast.selectiveFields ?? []) {
+    referenced.add(field.alias);
+  }
+  return lateMaterializedNodeAliases(ast).filter((alias) =>
+    referenced.has(alias),
+  );
+}
+
+type BuildLateMaterializedTopKCteInput = Readonly<{
+  ast: QueryAst;
+  collapsedTraversalCteAlias?: string;
+  dialect: DialectAdapter;
+  fromClause: SQL;
+  limit: number;
+  offset?: number | undefined;
+}>;
+
+/**
+ * The topK CTE: selects each node alias's identity `(id, kind)` plus each sort
+ * key value (aliased `__lm_sk{n}` so the outer SELECT can re-order the survivors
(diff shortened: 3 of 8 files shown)


pdlug:
## Review + fixes applied (commit 732942f9)

Adversarial review found one blocking bug (reproduced, not just theorized) plus two smaller items; all fixed directly on this branch. The PR description has been rewritten to match the current state (it feeds the squash merge).

### P1 (reproduced): recorded-time reads returned duplicated rows with stale-version props

The outer re-fetch joins the node source on `(graph_id, kind, id)` with **no temporal predicate**, on the assumption that this is "a pure PK re-fetch." That holds on the live table — that key is its PK — but under a recorded pin (`store.asOfRecorded(rt)`), `recordedReadSchemaFor` swaps `ctx.schema.nodesTable` to `typegraph_recorded_nodes`, where that key matches **one row per recorded version**. The gate never checked `ast.recordedAsOf`, so recorded selective `ORDER BY … LIMIT` queries reached this branch.

Reproduction (3 Person nodes, each with 2 recorded versions; `asOfRecorded(pin) … select({name, email}) .orderBy(age desc) .limit(2)`): **4 rows returned instead of 2** — each survivor duplicated, and with post-pin `email` values leaking into the pinned read where versions differed.

**Fix:** recorded reads (`ast.recordedAsOf !== undefined`) now fall back to the flat plan — consistent with this PR's existing out-of-envelope gating (the flat plan applies the temporal filter everywhere and is correct at any coordinate). A new case in the shared integration suite builds multi-version history and asserts exact single-version rows; it was verified to fail with the duplicated 4-row result when the gate is disabled. The `buildLateMaterializedOuterFromClause` comment now states the live-PK-uniqueness assumption explicitly.

### P2: "result-identical to the flat plan" was overstated

`buildStandardOrderBy` emits no total-order tiebreak, so when ties span the LIMIT boundary both the flat plan and the topK pick an arbitrary tied subset — and being different plan shapes, they may pick *different* tied rows. This is the flat plan's existing non-determinism class, not corruption, but the test-suite header claimed identity unconditionally. The claim is now scoped to unique top-Ks, with the tie-boundary caveat documented. (Adding an `id` tiebreak to `buildStandardOrderBy` would change flat-plan behavior too and is a separate decision — happy to file it.)

### P3: third inline copy of the candidate-CTE prefix

The start + traversal CTE loop was re-inlined here alongside the flat path's copy. Extracted `buildStandardStartAndTraversalCtes` and pointed both call sites at it — the late-mat plan feeds it the lean AST/columns and no per-traversal limit, the flat plan its originals.

### Confirmed sound during review (no changes needed)

- Live-table + valid-time `asOf` correctness: the re-join re-fetches exactly the row the topK validated (PK uniqueness); soft-deleted rows never enter the topK.
- #248 template-cache / reserved read-instant placeholder: the instant is bound inside the inner CTEs via the shared temporal pass; the outer join never references it.
- Gating: edge-field projections/orders, optional traversals, aggregates, vector/fulltext, `LIMIT <= 0` all fall back correctly; order-only columns stay in the lean CTE (`collectRequiredColumnsByAlias`'s orderBy marking is outside the new `includeProjection` guard); subclass expansion is safe (join keys on the concrete carried kind).
- `stripLimitOffsetFromPlan` mirrors the count fast path and feeds only the emitter's shape assertion.

### Verification

- Full SQLite suite: 5,022 passed (includes the 9-case late-mat suite on SQLite + libSQL + PGlite).
- PostgreSQL lane: 71 files / 2,072 tests passed.
- `pnpm fix` + `pnpm typecheck` clean.

PR #284, opened 2026-07-16
feat(indexes): system-index declarations with boot materialization

pdlug:
TypeGraph's base-relation indexes were maintained as two hand-synced lists (one per dialect schema file) and applied only by first-boot bootstrap DDL. That structure produced the exact failure #282 exposed: an index added in a newer library version **never reaches an already-initialized database** — `bootstrapTables()` only runs on a missing-table error, `createStore`/`createVerifiedStore` are zero-DDL by contract, and `schema/migration.ts` classifies index diffs as "safe" but never emits index actions. With six index-touching perf PRs merged in three weeks and more coming from the benchmark work, the per-index bolt-on pattern needed a foundation.

## Design

Complete the existing seam rather than adding a new abstraction: graph-declared indexes already have declarations, a DDL generator, a materializer with durable status rows, drift signatures, and a Postgres `CREATE INDEX CONCURRENTLY` cross-caller claim protocol. This PR brings the system indexes into that machinery.

- **`SYSTEM_INDEX_DECLARATIONS` — single source of truth** (`src/indexes/system.ts`) for the 21 plain btree indexes on nodes/edges/recorded relations, with per-table column-union types and the access-path rationale comments consolidated in one place. Both dialect schema factories derive their Drizzle index builders from it (columns resolved by each drizzle column's own physical `.name` — no naming convention in between), so **the two dialects cannot drift by construction**. Generated DDL is byte-identical to main: same names, columns, and order.
- **`materializeSystemIndexes()`** rides the shared materialization frame (status rows, drift signatures, invalid-leftover healing, CIC claim protocol) with system-specific physical-state semantics: a system index is fully identified by its library-owned name, so one bulk catalog read (scoped to the session `search_path` via `pg_table_is_visible`) both settles present indexes with zero DDL/writes **and** forces a rebuild when an index is physically absent or INVALID despite a stale success row (dump/restore, manual drop). Warm call: two concurrent reads, no DDL.
- **Boot integration**: `createStoreWithSchema` brings system indexes up to the running library version (the same privileged-boot step as fulltext/vector contributions). Lenient by contract — capability-missing backends are skipped, per-index and infrastructure failures degrade to a warning, the store always comes up. `systemIndexes: "skip"` opts out for deployments that must not run index builds inline at boot; **`store.materializeSystemIndexes()`** is the strict explicit API for zero-DDL deployments (replaces #282's "call `backend.bootstrapTables()`" guidance).
- **Plan-contract helpers**: `createPlanCaptureBackend()`/`explainQueryPlan()` extracted into `tests/test-utils.ts` so index changes can assert actual `EXPLAIN QUERY PLAN` shapes, not just DDL presence (the `degree-index-alignment` tests now consume them).

## Review hardening

The branch went through a 4-agent quality pass and a 19-agent high-effort correctness review; all verified findings are addressed in the third commit, including three that reached beyond the new code:

- The Postgres catalog probes are `search_path`-scoped — a schema-per-tenant database can never settle (or steer the `DROP INDEX` heal at) another schema's identically-named index.
- `buildMaterializationOnConflictSet` no longer overwrites a status row's shape columns on a **failed** attempt — signature drift now reports persistently instead of self-silencing after one run (pre-existing frame bug; fixes relational drift reporting too).
- A status row owned by a graph-declared index that collides with a system index name is refused **without writing**, instead of drift-failing over it and permanently bricking the user's index row.

Documented limitations (deliberate, in `system.ts`): an in-place declaration reshape is undetectable on bootstrap-created databases — the rename-on-reshape rule is enforced by review and the parity tests; and a renamed suffix strands the old physical index until a retired-suffixes drop path is added alongside the first real rename.

## Tests

- Cross-dialect parity: extraction test pins both generated DDL scripts' full index sets (names, columns, uniqueness) to each other; declaration-coverage test on both dialects.
- Upgrade paths: runtime adoption + boot adoption on SQLite; CONCURRENTLY-path adoption with status-row assertions on PostgreSQL.
- Hardening regressions: rebuild-on-physically-missing, persistent drift reporting, collision-refusal with untouched foreign row.
- Full SQLite suite 5,031 passed; PostgreSQL lane 71 files / 2,074 passed; `pnpm fix`, `pnpm typecheck`, knip clean.

Changeset: minor (`store.materializeSystemIndexes()`, `systemIndexes` boot option, `SYSTEM_INDEX_DECLARATIONS` export, `IndexEntity` gains `"system"`).

### DIFF
--- a/.changeset/system-index-declarations.md
+++ b/.changeset/system-index-declarations.md
@@ -36,5 +36,14 @@ gap #282 exposed). Now:
 - `IndexEntity` gains a `"system"` member; system status rows carry the
   relation key (e.g. `"recordedNodes"`) in their `kind` column.
 
-Generated DDL is unchanged — same index names, columns, and order — so
-existing databases and drizzle-kit migrations are unaffected.
+Generated DDL is unchanged for default and short custom table names — same
+index names, columns, and order — so existing databases and drizzle-kit
+migrations are unaffected. Names that would exceed PostgreSQL's 63-char
+identifier bound (very long custom table names) are now deterministically
+truncated + hash-suffixed instead of being silently truncated by the engine
+into collisions. System index names are reserved: a graph-declared index
+using one is rejected at table definition and by `materializeIndexes()`
+(previously its `CREATE INDEX IF NOT EXISTS` silently no-opped against the
+differently-shaped system index while recording success). Legacy databases
+that predate the recorded relations skip those indexes cleanly instead of
+attempting failing DDL at every boot.
--- a/.changeset/system-index-declarations.md
+++ b/.changeset/system-index-declarations.md
@@ -20,8 +20,15 @@ gap #282 exposed). Now:
   CONCURRENTLY` on PostgreSQL, riding the same status table, drift
   signatures, invalid-leftover healing, and cross-caller claim protocol as
   graph-declared indexes. A database whose indexes all exist settles from
-  one catalog read with no DDL and no writes. Failures degrade to a warning
-  (indexes are a performance concern; the store still boots).
+  one catalog read (scoped to the session `search_path`, so schema-per-
+  tenant databases never observe each other's indexes) with no DDL and no
+  writes; a system index that is physically absent or invalid is rebuilt
+  even when a stale success row survives (dump/restore, manual drop).
+  Failures — including status-table infrastructure errors — degrade to a
+  warning: indexes are a performance concern and the store still boots.
+  Deployments that must not run index builds inline at boot pass
+  `systemIndexes: "skip"` to `createStoreWithSchema` and materialize
+  out-of-band.
 - **New API: `store.materializeSystemIndexes()`** for deployments that boot
   without `createStoreWithSchema` (zero-DDL attach) — call once under a
   DDL-capable role after upgrading. Strict where the boot path is lenient:
--- a/.changeset/system-index-declarations.md
+++ b/.changeset/system-index-declarations.md
@@ -1,0 +1,33 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+TypeGraph's base-relation indexes are now **system-index declarations** — a
+single declared list (`SYSTEM_INDEX_DECLARATIONS`) that both dialect schemas
+derive from and that materializes onto already-initialized databases.
+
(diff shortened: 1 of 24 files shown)


pdlug:
## Review findings — all three fixed (commit 02f85fa1)

**P1 — 63-byte identifier bound.** Confirmed, and worse than bootstrap breakage: the catalog probes compare by exact untruncated name, so a long custom table name would also desync the name-presence fast path from the engine-truncated stored name and re-attempt the build on every boot. `systemIndexName` now deterministically truncates + hash-suffixes past 63 chars (same pattern as the vector identifier bound, reusing its `shortHash`), and since the Drizzle builders, runtime DDL, catalog probes, and status keys all derive the name from that one function, they can never disagree. Applied on both dialects so parity holds; short/default names are byte-identical to before. Tests: within-table distinctness under a 58-char table name (the `kind_idx`/`kind_created_idx` collision pair), determinism, no-hash-for-short-names, and bounded names appearing in generated DDL.

**P1 — collision guard missed bootstrap-created system indexes.** Confirmed — the entity-row guard only covered the materializer-adopted path; steady-state system indexes have no row, so a same-named graph declaration would `CREATE INDEX IF NOT EXISTS` into a silent no-op and record a false success. System index names are now **reserved** at two layers: both schema factories throw at table-definition time (covers the drizzle-kit path, resolved against the deployment's actual table names), and `materializeIndexes` fails a colliding declaration per-entry with no DDL and no status write. Tests cover both layers, including that no false-success row lands.

**P2 — legacy databases without the recorded relations.** Confirmed (`operation-backend-core` guards its clears for exactly this population). `materializeSystemIndexes` now preflights target-relation existence — one `search_path`-scoped catalog read folded into the existing concurrent preload, so the warm-boot read count is unchanged — and reports `skipped` (reason: relation does not exist) for indexes whose relation the database predates. A legacy pre-recorded database boots quietly; if the relation appears later, bootstrap creates its indexes with it. Test drops the recorded tables and asserts skipped entries plus a quiet subsequent boot.

Verification: full SQLite suite 5,036 passed; PostgreSQL lane running now (will confirm); `pnpm fix`/typecheck/knip clean. The changeset documents the identifier bound, the reserved-name contract, and the legacy-relation skip.

PR #285, opened 2026-07-16
perf(algorithms): remove iterative-round overheads (WCC join, round-trips, work_mem)

pdlug:
The SF1 `EXPLAIN (ANALYZE, BUFFERS)` investigation against pgGraph-on-the-same-engine attributed tg-postgres's iterative-algorithm gap to three removable overheads (none of them missing indexes — every access path was already index-served):

1. **WCC propagate rounds re-validated node visibility per edge** — two Index-Only-Scan loops over `typegraph_nodes` at `loops=361,246` (~3.17M buffers/round at SF1), re-checking endpoints whose working-table membership already proves visibility + scope.
2. **Per-round bookkeeping ran as separate statements** — IC13-shape traversals issued ~11 sequential round-trips for 4.8ms of actual SQL (~50% round-trip latency): separate seed COUNT, separate bidirectional meeting probes, and 1–3 working-table ANALYZE statements per traversal.
3. **Sorts spilled to disk every round** — WCC's 722k-row candidate sort wrote `external merge Disk: 106128kB` per round; BFS `DISTINCT` spilled similarly, all at default `work_mem`.

## Changes (one commit per lever)

**1. WCC: drop the redundant per-edge node-visibility join.** The invariant was verified before removal: the working table is seeded once through the same graph/kind/temporal filters the join re-applied (the join was strictly weaker — it never checked `nodeKinds`), inside one repeatable-read transaction with a single bound read instant, and WCC rounds only UPDATE label columns — they never insert. Membership ⇒ visibility + scope, so the join proved nothing. `propagateChunkLabels` now uses the edge-only expansion (`compileWorkingTableEdgeExpansion`) that reachable rounds already had. `reachable`/`shortestPath` rounds are untouched — they insert new nodes each round, so their target-visibility join is load-bearing. A unit test pins the round SQL shape (no `typegraph_nodes` join in propagate).

**2. Fold per-round bookkeeping into the round statements.** Seeds are `INSERT … RETURNING` (frontier count + seed-level meeting come back with the seed). Bidirectional rounds detect meetings via a new `meeting_depth` working-table column populated by a scalar subquery inside the expansion's INSERT and carried out through RETURNING — the only cross-dialect single-statement shape (SQLite forbids both wCTEs and subqueries in RETURNING). Correctness: working-table rows are immutable once inserted, and any meeting among pre-existing rows would have converged an earlier round, so newly inserted rows are a complete probe. On the 3-hop regression fixture: **shortestPath 14 → 8 statements, reachable 8 → 7**, pinned by budget tests (upper bounds + shape assertions: no `SELECT COUNT`, no standalone meeting probe). ANALYZE thresholds are unchanged but no longer emitted after the final budgeted round. `supportsTemporaryIteration` now also requires `capabilities.returning !== false` (the temp-table path already depended on RETURNING; a `returning: false` custom backend would have broken on main and now takes the inline fallback).

Equal-depth meeting **tie-breaks** are now UTF-16 code-unit order in JS — deterministic and identical across backends (consistent with WCC's representative ordering), where the old SQL `ORDER BY` was database-collation-dependent. Called out in the changeset: a linguistically-collated PG cluster may pick a different (equally shortest) path than the previous release when multiple meetings tie.

**3. Opt-in transaction-scoped working memory.** New `DialectAdapter.setTransactionWorkingMemory` seam (the `analyzeTemporaryTable` pattern): PostgreSQL emits parameterized `SELECT set_config('work_mem', $1, true)` — `SET LOCAL` semantics, transaction-scoped, reverts on commit/abort, session and server never touched, injection impossible by construction; SQLite no-ops. Exposed as `workingMemory?: string` on the iterative algorithm options (`reachable`/`shortestPath`/`canReach`/`neighbors`/`weaklyConnectedComponents`), validated to PostgreSQL's actual `work_mem` range (64kB–2147483647kB) with the same `ConfigurationError` on both backends. **The default inherits the server's configured `work_mem`** — `work_mem` is a per-sort/hash-operator (and per parallel worker) threshold, not a per-operation budget, so a library default would silently override the DBA's setting and multiply under concurrency (post-review fix). Set it explicitly (e.g. `"64MB"`) for large single-tenant analytical runs; the SF1 bench harness should do so to realize the spill elimination. The inline (non-transactional) fallback never emits it.

## Expected impact (SF1, to be confirmed by the bench re-run)

- GA_WCC on PG: ~48–65s → est. 15–25s (join elimination + no disk spills).
- IC13-lane: ~2× on the round-trip-bound constant factor.
- All graph algorithms on PG: no more per-round 106MB disk spills when `workingMemory` is set for the run.

These shave removable overhead; they do not close the architectural gap to pgGraph's in-memory CSR (single C call vs SQL iteration) — that framing is per the investigation report.

### DIFF
--- a/.changeset/iterative-algorithm-overheads.md
+++ b/.changeset/iterative-algorithm-overheads.md
@@ -27,11 +27,15 @@ collation could pick a different (equally shortest) path.
 
 New option: iterative algorithm calls (`reachable`, `shortestPath`,
 `canReach`, `neighbors`, `weaklyConnectedComponents`) accept
-`workingMemory?: string` (default `"64MB"`), a transaction-scoped memory
-budget applied on PostgreSQL with `SET LOCAL work_mem` semantics via
-parameterized `set_config`. It prevents whole-graph rounds from spilling
-their sorts to disk (measured ~106MB external merges per WCC round on SF1),
-never touches the session or server setting, is validated as
-`<digits>kB|MB|GB` within PostgreSQL's accepted `work_mem` range
-(64kB–2147483647kB) with the same typed error on both backends, and is
-ignored by SQLite.
+`workingMemory?: string`, an opt-in, transaction-scoped override of the
+session's `work_mem`, applied on PostgreSQL with `SET LOCAL` semantics via
+parameterized `set_config`. By default (option omitted) operations inherit
+the server's configured `work_mem` — nothing is overridden. `work_mem` is a
+threshold each sort/hash operator (and each parallel worker) may allocate up
+to, not a per-operation budget, and concurrent calls multiply it; set it
+deliberately (e.g. `"64MB"`) for large single-tenant analytical runs where
+the configured default spills whole-graph sorts to disk (measured ~106MB
+external merges per WCC round on SF1). The override never touches the
+session or server setting, is validated as `<digits>kB|MB|GB` within
+PostgreSQL's accepted `work_mem` range (64kB–2147483647kB) with the same
+typed error on both backends, and is ignored by SQLite.
--- a/.changeset/iterative-algorithm-overheads.md
+++ b/.changeset/iterative-algorithm-overheads.md
@@ -20,6 +20,10 @@ shortest-path traversal that used to issue two to three statements per round
 now issues one, roughly halving round-trip latency on latency-bound
 connections. The working-table `ANALYZE` policy is unchanged in its
 thresholds but no longer runs when no further round will read the table.
+When several equal-depth meetings exist, the tie now breaks by node id then
+kind in code-unit order on both backends — previously the selection followed
+the database collation, so a PostgreSQL cluster with a linguistic default
+collation could pick a different (equally shortest) path.
 
 New option: iterative algorithm calls (`reachable`, `shortestPath`,
 `canReach`, `neighbors`, `weaklyConnectedComponents`) accept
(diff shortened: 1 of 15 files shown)


pdlug:
## Review response (commits 2659c307, 6b3ff719)

**Unused-code check** — fixed in 2659c307: `IterativeMemoryOptions` is only consumed inside `algorithms/types.ts` (the composed option types intersect it), so the export is dropped; the public surface exposes the composed types.

**P1 work_mem default** — agreed and fixed in 6b3ff719. The default now **inherits the server's configured `work_mem`**: no `set_config` is emitted unless the caller passes `workingMemory` explicitly. `work_mem` is a per-sort/hash-operator (and per parallel worker) threshold, not a per-operation budget, so a library-wide 64MB default would have silently overridden the DBA's setting and multiplied under concurrency. Docs/JSDoc/changeset rewritten with the correct semantics and an explicit-opt-in recommendation for large single-tenant analytical runs. New PG wiring tests prove zero `set_config` statements when the option is omitted and exactly one parameterized override (emitted before `CREATE TEMP TABLE`) when it's passed.

**Bench note:** the SF1 re-run should pass `workingMemory` (e.g. `"64MB"`) on the tg-postgres algorithm calls to realize the spill elimination — the library no longer does it implicitly.

Lanes after the fix: SQLite 5,059 passed; PostgreSQL 72 files / 2,092 passed. PR description updated to match.

PR #288, opened 2026-07-17
feat(algorithms): weighted shortest path (LDBC IC14)

### LINKED ISSUE #275: store.algorithms: weighted shortest path (LDBC IC14)
Split from #272. `store.algorithms.shortestPath` (set-based bidirectional BFS as of #274) is unweighted. LDBC Interactive IC14 needs a **weighted** shortest path (e.g. path weighted by interaction count). Add a weighted variant so IC14 can be expressed faithfully rather than approximated by the unweighted IC13. Relates to the traversal work in #274; benchmark would add a weighted-path lane once available.

pdlug:
Adds `store.algorithms.weightedShortestPath(from, to, options)` — a minimum-total-weight path search where each traversed edge contributes a numeric edge property (`weightProperty`) to the path cost. This is the shape LDBC Interactive IC14 needs: the cheapest route by interaction weight, which is usually not the fewest-hop route.

## API

```typescript
const path = await store.algorithms.weightedShortestPath(alice, bob, {
  edges: ["knows"],
  weightProperty: "interactionCost",
  direction: "both",
});
// { nodes: [...], depth: 3, totalWeight: 2.5 } | undefined
```

- Separate method rather than a `weight` option on `shortestPath`, per the convention that option flags must not change return types.
- Options mirror the sibling algorithms (`edges`, `direction`, temporal options, `workingMemory`); `maxIterations` (default 1000, like WCC) replaces `maxHops` — cost-ordered search doesn't settle nodes in hop order, so the round cap is a runaway backstop, not a depth bound.
- Exposed on `Store`, `StoreView` (`store.asOf(...).weightedShortestPath`), and recorded views, with the same `recordedAsOf` gating as the other algorithms.

## Algorithm

Frontier-based label-correcting relaxation on the #274 iterative substrate: each round relaxes edges out of the nodes improved in the previous round, keeping one best `(distance, predecessor)` per `(kind, id)` identity, until no distance improves. Candidates that already cost at least as much as a known path to the target are pruned (sound because weights are non-negative), with the bound tracked in JS from each round's `RETURNING` rows so pruning is a plain bind parameter.

Both substrate execution paths are implemented, mirroring BFS: temporary working-table rounds on transactional backends, and the chunked inline `VALUES` fallback elsewhere. Self paths short-circuit to a visibility check plus the weight audit.

## Weight contract

Weights are audited **up front**, inside the operation's snapshot, across every visible edge of the selected kinds — so a weighted call either observes a fully valid weight domain or fails fast with the new typed `InvalidEdgeWeightError` naming the offending edge:

- `negative` — target pruning assumes non-negative weights;
- `non_numeric` — the property must be a JSON number (a JSON string `"5"` or `"null"` does not count);
- `not_finite` — a JSON number beyond the IEEE 754 double range (checked in the NUMERIC domain on PostgreSQL, so it reports the typed error instead of a raw 22003 overflow);
- `missing` — unless `defaultWeight` is configured, in which case it substitutes.

The global scan is a deliberate O(E-of-selected-kinds) cost per call: reachability-scoped validation would make data errors nondeterministic in which endpoints you happen to query.

## Backend parity

- Weight arithmetic is IEEE 754 double on both backends via a new `DialectAdapter.jsonExtractDouble` seam (`json_extract` / `::double precision`) — `jsonExtractNumber`'s PostgreSQL `::numeric` would accumulate in exact decimal and diverge from SQLite doubles.
- Two new type-based dialect predicates, `jsonPathIsNumber` and `jsonPathIsMissingOrNull`, are never-NULL by contract; the latter exists because `jsonPathIsNull`'s PostgreSQL text-comparison form goes three-valued on a JSON `null` and misclassifies the JSON string `"null"`.
- Equal-weight tie-breaks (predecessor choice, multi-kind target selection) use code-point comparisons matching SQLite `BINARY` / PostgreSQL `COLLATE "C"` on both execution paths, and equal-distance candidates carrying the target id survive pruning so multi-kind target ids honor the documented smallest-identity tie-break.
- Documented corner: edge-kind lists large enough to exceed the bind-parameter budget (hundreds of kinds in one call) can resolve equal-weight predecessor ties differently across backends; totals are unaffected.

## Validation

- New coverage: cross-backend integration tests (min-weight vs min-hop, `direction: "both"`, missing/default weights, JSON null vs string `"null"`, fractional double-parity, temporal + pinned StoreView) and unit tests (parallel edges, zero-weight cycles, self path, direction variants, inline-fallback parity, convergence error, multi-kind target tie-break, out-of-range weight injection, option validation)

Closes #275

### DIFF
--- a/.changeset/weighted-shortest-path.md
+++ b/.changeset/weighted-shortest-path.md
@@ -1,0 +1,14 @@
+---
+"@nicia-ai/typegraph": minor
+---
+
+Add `store.algorithms.weightedShortestPath` — a minimum-total-weight path
+search weighting each traversed edge by a numeric edge property (LDBC
+Interactive IC14 shape). Runs frontier-based relaxation on the shared
+iterative substrate with best-target pruning, works on both execution paths
+(temporary working table and inline fallback), and honors valid-time and
+recorded-time coordinates including pinned StoreViews. Edge weights are
+audited up front: negative, non-numeric, or (without `defaultWeight`)
+missing weights throw the new typed `InvalidEdgeWeightError`. Weight
+arithmetic is IEEE 754 double precision on both backends, so paths and
+totals are backend-identical.
--- a/.changeset/weighted-shortest-path.md
+++ b/.changeset/weighted-shortest-path.md
@@ -8,7 +8,10 @@ Interactive IC14 shape). Runs frontier-based relaxation on the shared
 iterative substrate with best-target pruning, works on both execution paths
 (temporary working table and inline fallback), and honors valid-time and
 recorded-time coordinates including pinned StoreViews. Edge weights are
-audited up front: negative, non-numeric, or (without `defaultWeight`)
-missing weights throw the new typed `InvalidEdgeWeightError`. Weight
-arithmetic is IEEE 754 double precision on both backends, so paths and
-totals are backend-identical.
+audited up front: negative, non-numeric, out-of-range, or (without
+`defaultWeight`) missing weights throw the new typed
+`InvalidEdgeWeightError`. Weight arithmetic is IEEE 754 double precision on
+both backends, so total weights are backend-identical; among
+equal-total-weight paths the returned node sequence is too, except when the
+`edges` list exceeds the backend's bind-parameter budget (hundreds of edge
+kinds in one call).
--- a/README.md
+++ b/README.md
@@ -42,10 +42,11 @@ TypeGraph works well for:
 TypeGraph is not designed for:
 
 - Distributed graph processing
-- Advanced graph analytics (PageRank, community detection, weighted shortest
-  path, centrality beyond degree) — it does ship connectivity algorithms
-  ([Graph Algorithms](https://typegraph.dev/graph-algorithms): unweighted
-  shortest path, reachability, neighborhoods, degree, exact WCC)
+- Advanced graph analytics (PageRank, community detection, centrality
+  beyond degree) — it does ship connectivity algorithms
+  ([Graph Algorithms](https://typegraph.dev/graph-algorithms): shortest
+  path both weighted and unweighted, reachability, neighborhoods, degree,
+  exact WCC)
 - Billion-scale graphs requiring dedicated graph-engine performance
 
 ## Installation
--- a/apps/docs/src/content/docs/graph-algorithms.md
+++ b/apps/docs/src/content/docs/graph-algorithms.md
@@ -137,7 +137,7 @@ Options differ from the unweighted traversals:
 |--------|------|---------|-------------|
 | `edges` | `readonly EdgeKinds<G>[]` | *(required)* | Edge kinds to follow |
 | `weightProperty` | `string` | *(required)* | Top-level edge property holding each edge's non-negative numeric weight |
-| `defaultWeight` | `number` | *(none)* | Substituted for edges missing the property; must be finite and non-negative |
+| `defaultWeight` | `number` | *(none)* | Substituted for edges missing the property; must be non-negative and within the audit's upper bound (~9.7e289) |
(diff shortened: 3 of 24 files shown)


pdlug:
All ten findings addressed in 9cd3243:

**[P1] Equal-bound pruning vs zero-weight ties** — fixed by pruning only strictly-worse candidates (`<=` bound admits the equal-cost plateau) and deleting the target-id carve-out, which this generalizes. Termination is unaffected: the upsert still accepts only strict improvements, so the plateau is finite. Regression test drives a zero-weight edge from an equal-cost intermediate to the smaller-identity target kind, on both execution paths.

**[P1] Weight-domain divergence (underflow / accumulated overflow)** — the audit now bounds the accepted domain: magnitudes above `Number.MAX_VALUE / 2^64` (~9.7e289) and nonzero magnitudes below `5e-324` throw the typed error (reason renamed `not_finite` → `out_of_range`). The upper bound makes accumulated overflow *provably* impossible — a path has at most 2^53 hops (one row per hop), and 2^53 · bound stays ~2000x below the double max even after addition rounding — so no per-candidate guard is needed on the hot path. One declared engine caveat: SQLite's JSON parser rounds sub-denormal text like `1e-400` to `0` before SQL can observe it, so only PostgreSQL can reject that case; documented on the constant and in the docs page.

**[P1] Raw literal escaping** — `escapePostgresLiteral` now emits the configuration-independent `E'…'` form (backslashes and quotes doubled) whenever a value contains a backslash; backslash-free values keep the plain form so emitted SQL text is unchanged for the common case, preserving cross-clause expression-text matching.

**[P2] Frontier index** — added `(graph_id, run_id, improved_round)` temp index at initialize.

**[P2] Absent target** — both paths now check target visibility after the audit and return `undefined` without relaxing (working-table path converges with an empty table; inline checks both endpoints like BFS).

**[P2] Result extraction** — replaced the full-table read with a recursive CTE that picks the cheapest target row (kind tie in binary collation) and walks predecessor pointers; transfer is O(path length), with a `position` guard bounding recursion at the round budget.

**[P2] Typed details** — `InvalidEdgeWeightErrorDetails` exported and `declare`d on the class, matching the other typed error classes.

**[P2] Recorded reconstruction** — cross-backend test added: a cheaper detour appears after a recorded pin; the pinned view keeps returning the historical weight-5 direct path, the later pin the weight-2 detour, and replaying the earlier pin is stable.

**[P2] README + changeset/type-doc claims** — README updated; the backend-identical claim is now stated precisely (totals always identical; node sequences identical except when the `edges` list exceeds the bind-parameter budget).

Validation: `pnpm fix`, typecheck, full SQLite suite (5,116 passed), full PostgreSQL suite (2,120 passed).

pdlug:
Remaining P1 and the doc drift addressed in 61b58a9:

**defaultWeight bound** — `assertWeightOptions` now enforces the same upper bound the audit applies to stored weights (`> MAX_EDGE_WEIGHT` is rejected as a `ConfigurationError` before any SQL runs), closing the missing-weight overflow gap. Regression test covers both sides over missing-weight edges: `defaultWeight: Number.MAX_VALUE` rejects up front, and the exact bound is accepted with a finite two-hop sum. (No lower bound needed: a JS `number` is already a representable double and binds as a float8 parameter, so the sub-denormal cast-underflow path doesn't exist for defaults.)

**Doc drift** — the algorithm JSDoc and the docs page now say strictly-more pruning with equal-cost candidates explicitly retained, and the backend-parity paragraphs state the claim with the bind-budget qualification inline ("total weights are always backend-identical, and — unless the edges list exceeds the bind-parameter budget — so is the returned node sequence") instead of an unconditional sentence followed by an exception.

#290 rebased onto this and force-pushed. Full suites re-running; will confirm.

Issue #287, opened 2026-07-17
[Bug]: DO SQLite: history-mode recorded INSERT exceeds the 100 bound-parameter limit (backend advertises 999 in do-sqlite mode)


pdlug:
### Environment

- `@nicia-ai/typegraph` 0.35.0
- Backend: `createSqliteBackend(drizzle(ctx.storage))` (`drizzle-orm/durable-sqlite`, auto-detected `transactionMode: "do-sqlite"`)
- Runtime: workerd via `wrangler dev` / `@cloudflare/vitest-pool-workers`
- Store: `createStoreWithSchema(graph, backend, { history: true })`

### Symptom

With `history: true` on a Durable Object, a single `store.transaction()` that creates **≥ 7 edges** fails at commit with:

```
too many SQL variables: SQLITE_ERROR
```

The failure comes from the recorded-history capture flush: each committed transaction writes its after-images into `typegraph_recorded_edges` as one multi-row `INSERT` (~16 bound params per edge row), so 7 edges ≈ 112 params. Cloudflare caps SQLite-backed DOs at **100 bound parameters per query** ([Durable Objects limits](https://developers.cloudflare.com/durable-objects/platform/limits/) — "Maximum bound parameters per query: 100"). Empirically consistent: 5 edges (~80 params) per transaction commits fine, 7 fails.

### Root cause

The chunking machinery already exists and works — it is just fed the wrong budget on DO:

- `store/recorded-capture/relations.ts` — `recordedBindParamBudget()` honors `capabilities.maxBindParameters` (fallback 900), and `recordedChunkSize()` divides it by the column count. Correct behavior, correct design.
- `backend/types.ts:2237` — `SQLITE_MAX_BIND_PARAMETERS = 999`, and the SQLite backend capabilities advertise it unconditionally (`backend/types.ts:2267`).
- The do-sqlite execution profile is detected (`backend/drizzle/sqlite.ts` `transactionMode: "do-sqlite"`), but capabilities still claim 999 → recorded chunks of up to ~62 edge rows → ~992 params in one statement → hard failure on any transaction whose recorded rows exceed ~6 edges.

The same wrong budget also flows into the other batch sizers derived from `maxBindParameters` (`backend/drizzle/sqlite.ts:244-269` — unique checks, embedding/fulltext upserts, edge inserts), so bulk writes on DO presumably hit the same wall. Related: #22 covered this error class for bulk `getOrCreate` on D1.

### Suggested fix

[...]

PR #316, opened 2026-07-26
Fix PGlite bind-parameter batching

### LINKED ISSUE #315: [Bug]: PGlite ≥32,768-bind statements silently poison the connection; batching ignores backend limits
## Summary

`bulkCreate` can send a PostgreSQL statement with more bind parameters than the active backend supports. With PGlite 0.5.x, a statement containing **32,768 or more binds** resolves without an error but returns no rows; subsequent queries on the same client also return empty results.

This is narrower than the original report: real PostgreSQL accepts up to 65,535 binds and fails loudly at 65,536 without poisoning the connection. The PGlite boundary is 32,767.

## Verified behavior

| Backend / statement | Result | Connection afterward |
| --- | --- | --- |
| PGlite, 32,767 binds | succeeds | healthy |
| PGlite, 32,768 binds | silently returns no rows | subsequent `SELECT 1` returns no rows |
| PostgreSQL 18, 65,535 binds | succeeds | healthy |
| PostgreSQL 18, 65,536 binds | throws `08P01` | healthy |

TypeGraph reproductions:

- History-enabled edge `bulkCreate` with 2,048 rows emits a recorded-edge insert with `2,048 × 16 = 32,768` binds and poisons the PGlite client.
- History-disabled edge `bulkCreate` with 2,979 rows emits a live-edge insert with `2,979 × 11 = 32,769` binds and does the same.
- At one row below each boundary, the client remains healthy.

The writes are durable in the tested history-enabled case: reopening a clone shows all 2,048 live and recorded edges. The immediate correctness failure is that the original connection silently produces false-empty reads after the oversized statement.

## Root cause

There are three cooperating defects:

1. PGlite inherits the generic PostgreSQL capability value of 65,535 bind parameters, although its effective safe maximum is 32,767.
2. Most PostgreSQL batch sizes are computed once from the generic PostgreSQL maximum instead of from `backend.capabilities.maxBindParameters`. (Embedding batching already honors the runtime capability.)
3. The execution path has no final bind-count guard, so a future bad batch calculation or arbitrary oversized statement can still reach PGlite.

Schema width is not the cause for the core node/edge inserts: properties are serialized into a single JSON bind. The fixed estimates are conservative for those statements, but they are calibrated against the wrong backend maximum and ignore runtime capability overrides.

## Expected behavior

- PGlite advertises a safe `maxBindParameters` of 32,767.
- Every PostgreSQL batch size is derived from the active backend capability.
- Statements exceeding the advertised limit fail locally with a clear error before dispatch.
- Large node/edge writes are chunked without changing atomicity or history behavior.
- PGlite remains queryable after the boundary regressions above.
- Real PostgreSQL keeps its 65,535-bind capability.

## Reproduction outline

```ts
const rows = Array.from({ length: 2_048 }, (_, index) => ({
  fromId,
  toId,
  properties: { index },
}));

await store.edges.Relation.bulkCreate(rows);

// Before the fix, both reads silently appear empty/undefined on this client.
await client.query("SELECT 1 AS value");
await store.recordedNow();
```

## Fix scope

- Detect PGlite when constructing a PostgreSQL backend and clamp its advertised maximum to 32,767.
- Compute all PostgreSQL chunk sizes from the normalized runtime capability.
- Add a pre-dispatch bind-count guard.
- Add regressions for both history-enabled recorded-edge batching and history-disabled live-edge batching.

## Workaround

Keep every emitted PGlite statement below 32,768 binds by manually splitting bulk operations into smaller batches. For edges, a conservative batch of 2,000 rows stays below the observed limit for both live and recorded history inserts.

pdlug:
- detect PGlite-backed Drizzle databases and clamp their advertised bind-parameter budget to 32,767
- derive every PostgreSQL batch/chunk size from the active backend capability
- reject over-budget compiled and prepared statements before driver dispatch
- cover the live-edge and recorded-edge PGlite failure boundaries

## Root cause

PGlite 0.5.x silently returns no rows and leaves the current connection producing false-empty reads when a statement reaches 32,768 binds. TypeGraph advertised PostgreSQL's 65,535-bind ceiling for PGlite and computed most PostgreSQL batch sizes from that static value, so large edge batches could cross PGlite's lower limit.

The fix keeps real PostgreSQL at 65,535, clamps PGlite to 32,767, makes batching capability-driven, and adds a final execution guard so future sizing mistakes fail loudly before reaching the driver.

## Impact

Large PGlite bulk writes are now split safely for both ordinary and history-enabled stores. The client remains queryable after the previously failing 2,979-live-edge and 2,048-recorded-edge cases.

Fixes #315

### DIFF
--- a/packages/typegraph/src/backend/drizzle/execution/postgres-execution.ts
+++ b/packages/typegraph/src/backend/drizzle/execution/postgres-execution.ts
@@ -6,6 +6,7 @@ import {
 } from "drizzle-orm/pg-core";
 import { type ExtractTablesWithRelations } from "drizzle-orm/relations";
 
+import { ConfigurationError } from "../../../errors";
 import { isSqlFragment } from "../../../query/sql-fragment";
 import { shouldForceCustomPlan } from "../../../query/sql-intent";
 import { getOrCreateLru } from "./lru";
@@ -156,6 +157,19 @@ function isPgliteClient(candidate: unknown): candidate is NodePgClient {
     hasFunctionProperty(candidate, "query") &&
     hasFunctionProperty(candidate, "dumpDataDir")
   );
+}
+
+/**
+ * PGlite 0.5.x serializes the wire-protocol parameter count as a signed
+ * 16-bit integer. At 32,768 binds it silently returns no rows and leaves the
+ * client returning false-empty results, so the advertised budget must stay at
+ * the largest signed 16-bit value.
+ */
+export const PGLITE_MAX_BIND_PARAMETERS = 32_767;
+
+/** Detects a root Drizzle database backed by PGlite without importing it. */
+export function isPgliteDatabase(db: AnyPgDatabase): boolean {
+  return isPgliteClient((db as PgClientCarrier).$client);
 }
 
 function transactionSession(
@@ -489,9 +503,11 @@ async function executeDrizzleQuery<TRow>(
 function createPgPreparedStatement(
   pgClient: PgQueryClient,
   sqlText: string,
+  maxBindParameters: number | undefined,
 ): PreparedSqlStatement {
   return {
     async execute<TRow>(params: readonly unknown[]): Promise<readonly TRow[]> {
+      assertWithinBindParameterLimit(params.length, maxBindParameters);
       const result = await pgClient.query(sqlText, params);
       return result.rows as readonly TRow[];
     },
@@ -531,6 +547,12 @@ export type PostgresExecutionAdapterOptions = Readonly<{
    */
   preparedStatementCacheMax?: number;
   /**
+   * Maximum parameters that one statement may dispatch. Backend factories
+   * pass their normalized capability here so every execution surface has a
+   * final fail-loud guard after batch sizing.
+   */
+  maxBindParameters?: number;
+  /**
    * @internal Resolve Drizzle's pinned transaction client. Trusted import uses
    * this on its private adapter; ordinary transaction backends deliberately do
    * not expose a new raw-execution surface that would change routing behavior.
@@ -545,6 +567,7 @@ export function createPostgresExecutionAdapter(
   const prepareStatements = options.prepareStatements ?? true;
   const cacheMax =
     options.preparedStatementCacheMax ?? DEFAULT_POSTGRES_STATEMENT_CACHE_MAX;
+  const maxBindParameters = options.maxBindParameters;
   const pgClient = resolvePgClient(
     db,
     prepareStatements,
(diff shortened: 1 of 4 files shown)


pdlug:
Verified this against the original failure. It fixes it, and your root-cause diagnosis is better than the one I filed — **my issue is wrong on the mechanism and I've corrected it.**

### PGlite's limit is 32,767, not 65,535

I confirmed independently, straight against PGlite with no TypeGraph in the loop:

```
32766 params -> alive
32767 params -> alive
32768 params -> DEAD
```

A signed 16-bit boundary, exactly as you describe. My issue asserted the ceiling was PostgreSQL's 65,535, so every "actual params/row" figure in that table was exactly 2× inflated (I divided by the wrong limit). The knock-on error was concluding that `NODE_INSERT_PARAM_COUNT = 9` / `EDGE_INSERT_PARAM_COUNT = 12` were undercounts. They aren't — the real costs are ~7 and ~11, so the divisors were conservative all along and the budget was the only thing wrong. Same shape as #287 after all, which I'd argued against.

### Verification

Built `f4e5dbd5` and ran the case that started this — the demo's real corpus, **deliberately unchunked**, which is the exact call shape that used to wedge:

```
trials       1101 -> recordedNow ok
publications 1918 -> recordedNow ok
edges        2110 -> recordedNow ok

connection: ALIVE | typegraph_nodes = 3019 | 6514 ms
```

Both boundaries named in the PR body pass, and so does a sweep I ran to look for any remaining wedge — 20-property schema, `history: true`, single unchunked `bulkCreate`:

| rows | nodes | edges |
|---|---|---|
| 500 – 12,000 | alive at every step | alive at every step |

No wedge anywhere, including 12,000 rows × 20 properties. I'd expected wide schemas and history-enabled edges to still break, on the theory that params/row scales with property count and the fixed divisors couldn't cover it. That prediction was wrong: the node insert and the recorded-capture insert are separately chunked, so my inferred "combined" params/row was never a real quantity.

### On the design

The three layers look right to me, and the third is the one I care about most:

1. Clamping the advertised budget on detection is the correct fix location — capability, not call site.
2. Making every chunk size capability-driven removes the class, not just this instance. The static `POSTGRES_*_BATCH_SIZE` constants were a trap that would have re-fired on the next backend with a different ceiling.
3. `assertWithinBindParameterLimit` before dispatch is what stops the *next* sizing mistake from being silent. Given the failure mode here — reads returning empty rather than throwing, which for `asOfRecorded` means confidently wrong history — a loud failure at the boundary is worth more than the correct batch size.

One thing possibly worth a follow-up, non-blocking: `recordedNow()` returning `undefined` is still overloaded between "no clock row yet" (legitimate on a fresh graph) and "the clock read returned nothing unexpectedly". That overload is what made this take a while to find — the documented `if (instant === undefined)` handling treats connection loss as a benign empty graph. With the guard in place it's much less reachable, so it may not be worth changing.

Correction to my own last line: I'm *keeping* the 250-row workaround chunking in my demo for now, since it runs on published 0.42.0. I'll drop it once a release with this fix lands.

PR #364, opened 2026-07-30
feat(algorithms): bound whole-graph result extraction

### LINKED ISSUE #360: feat: add bounded extraction for whole-graph algorithm results
## Problem

Whole-graph algorithms always buffer and return every visible node after the expensive computation finishes. On multi-million-node graphs, the result transfer and second full-size JavaScript mapping allocation can be material even when the caller only needs the highest-ranked nodes or sufficiently large components.

This was split from #300 after the 0.45 maintainer triage.

## Proposed surface

- Optional `topK` for PageRank and Personalized PageRank, applied in extraction SQL after deterministic score ordering.
- A separately specified component-size filter for WCC. Define its public name and boundary behavior before implementation; `minSize` is the current candidate.
- Preserve the existing result element shapes.

These options bound result materialization only. Documentation must not imply they reduce the graph computation itself.

## Acceptance

- Limits are validated consistently and use typed errors.
- SQL performs the filtering/limit before rows reach the driver.
- Default behavior and ordering are unchanged.
- SQLite and PostgreSQL share the same semantic tests, including ties and boundary values.

pdlug:
- add positive-safe-integer `topK` to PageRank and Personalized PageRank
- add inclusive positive-safe-integer `minComponentSize` to weakly connected components
- apply both bounds in extraction SQL, preserving full-graph computation, existing row shapes, defaults, and deterministic ordering
- document the result-materialization boundary and add shared SQLite/PostgreSQL semantic coverage for ties, inclusive boundaries, defaults, and invalid values

Closes #360.

### DIFF
--- a/.changeset/calm-graphs-bound.md
+++ b/.changeset/calm-graphs-bound.md
@@ -1,0 +1,6 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Add SQL-level `topK` extraction for PageRank and personalized PageRank, plus
+inclusive `minComponentSize` filtering for weakly connected components.
--- a/.changeset/calm-graphs-bound.md
+++ b/.changeset/calm-graphs-bound.md
@@ -1,6 +1,10 @@
 ---
-"@nicia-ai/typegraph": patch
+"@nicia-ai/typegraph": minor
 ---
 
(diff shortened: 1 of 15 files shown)


pdlug:
Temporarily closing and reopening to schedule CI after retargeting the stacked PR to main.

PR #536, opened 2026-08-22
fix(store): fence endpoint get-or-create reads against stale transports

### LINKED ISSUE #530: getOrCreateByEndpoints match-key reads run outside the write transaction — stale-read transports break read-own-writes, and the failure blames a phantom competing writer
## Summary

`getOrCreateByEndpoints`' match-key / existence reads execute **outside** the write plan's transaction. On a transport that caches query results and only bypasses its cache inside explicit transactions (concretely: Cloudflare Hyperdrive with query caching enabled, in front of PostgreSQL), this breaks read-own-writes across calls: the pre-insert existence probe SELECT and the later match-key lookup compile to **byte-identical SQL and parameters**, so the transport caches the probe's empty result and then serves that stale emptiness to every subsequent lookup until the cache TTL expires.

The convergence loop then spins against its own cached shadow and fails with a diagnostic that asserts facts about the database that are false:

```
getOrCreateByEndpoints for works_at between person "..." and company "..." lost its match key
to a competing writer 3 times without converging. A concurrent writer is repeatedly creating
and removing this edge; serialize those callers or retry the operation.
```

In the incident we traced, the edges relation contained **exactly one** matching edge row, created once, never updated, never soft- or hard-deleted — there was no concurrent writer of any kind. The "repeatedly creating and removing" claim was an artifact of the stale reads.

## Two requests

1. **Run the match-key/existence reads inside the write plan's transaction.** On transaction-mode poolers (pgbouncer transaction mode, Hyperdrive, etc.) an explicit transaction pins one backend connection and bypasses result caches, so read-own-writes would hold by construction. This is the same class of hardening as #428's move from check-then-insert to claim-backed writes — the reads that feed the claim decision are the remaining piece outside the fence.
2. **Widen the diagnostic.** When the loop exhausts its retries, the honest statement is "the read consistently disagrees with the write" — which has (at least) two causes: a genuinely interleaving writer, or a stale-read transport between the library and the database. Mentioning the second would have cut our diagnosis time from hours to minutes, and would stop the error from sending operators hunting for a writer that does not exist.

For completeness: we resolved the production occurrence by disabling the transport's query cache, and TypeGraph behaves correctly for us with the cache off. But the library's read-before-write pattern is what makes an otherwise-transparent caching pooler unsafe, and the pattern is invisible to consumers until it fires.

Observed with `@nicia-ai/typegraph` 0.50.0 on Cloudflare Workers → Hyperdrive (query caching enabled, since disabled) → PostgreSQL (Neon).

pdlug:
Make getOrCreateByEndpoints derive match decisions from transaction-owned reads and converge directly from the authoritative row found by the write fence.

This prevents caching transports from replaying stale empty or positive matches, protects update, resurrection, and bulk paths from stale candidates, and reports both concurrent-writer and stale-transaction-read causes when convergence cannot be established.

Closes #530

### DIFF
--- a/.changeset/transactional-edge-match-reads.md
+++ b/.changeset/transactional-edge-match-reads.md
@@ -1,0 +1,5 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Keep edge get-or-create match decisions inside their write transaction.
--- a/packages/typegraph/src/store/operations/edge-operations.ts
+++ b/packages/typegraph/src/store/operations/edge-operations.ts
@@ -88,8 +88,8 @@ import {
   type GraphReadBackend,
   type InsertEdgeParams,
   rowPropsToObject,
-  type TransactionBackend,
   runOptionallyInTransaction,
+  type TransactionBackend,
 } from "../../backend/types";
 import { validateEdgeEndpoints } from "../../constraints";
 import { type GraphDef } from "../../core/define-graph";
--- a/packages/typegraph/src/store/operations/edge-operations.ts
+++ b/packages/typegraph/src/store/operations/edge-operations.ts
@@ -89,6 +89,7 @@ import {
   type InsertEdgeParams,
   rowPropsToObject,
   type TransactionBackend,
+  runOptionallyInTransaction,
 } from "../../backend/types";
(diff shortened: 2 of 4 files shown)


pdlug:

Issue #533, opened 2026-08-22
Single managed write plans cost 5-6 round trips each — fence probe and duplicate checks dominate on high-RTT links


pdlug:
## Observation

With statement logging on the PostgreSQL adapter (prepared statements disabled), a **single** managed node or edge write costs 5–6 sequential statements on the connection:

```
begin
SELECT ... schema-version fence probe
SELECT ... duplicate/endpoint check
INSERT ... (the write)
commit
```

Each is one round trip. On a low-RTT link this is invisible; on edge-compute-to-managed-PostgreSQL links (~45ms RTT) every entity write costs ~250–300ms, and a flow that creates a handful of nodes and edges serially spends multiple seconds almost entirely in these fixed per-plan overheads. In a captured first-provisioning flow of one graph (97 statements total), the per-write fence/probe/tx framing accounted for roughly 40 statements across ~7 writes.

`store.batch()` helps a lot where it applies — `batch(q1..q4)` measures as `begin` + 4 + `commit` — and we are moving the flows we control onto it. This issue is about the residual costs that batch does not remove and single writes always pay:

## Asks / discussion

1. **Schema-version fence probe caching.** The fence probe re-reads the committed schema version on every plan. Could it be cached per (connection or backend instance, graph) and invalidated by the schema-evolution paths — or folded into the write statement itself (e.g. `WHERE` against the version row with a row-count check), turning probe+write into one statement?
2. **Duplicate/endpoint checks folded into the write.** For creates, the duplicate-id probe could ride the INSERT (`ON CONFLICT DO NOTHING` + row-count, or a CTE that selects-then-inserts) rather than being a separate round trip; endpoint existence for edge creates could similarly be a joined INSERT ... SELECT that fails typed when the endpoint rows are absent.
3. **Guidance:** if some of this is already achievable with existing options (beyond `batch()`), a "chatty-link deployment" section in the backend docs would help — the per-plan overhead is invisible until the RTT makes it the dominant term.

Measured with `@nicia-ai/typegraph` 0.50.0/0.51.0, bundled PostgreSQL backend over drizzle node-postgres, `prepareStatements: false`.

PR #538, opened 2026-08-22
perf(store): avoid identity locks for generated node ids

pdlug:
Skip identity advisory locks and cross-kind fold probes when TypeGraph generated the node id, because a fresh generated id cannot collide with an existing identity.

Keep caller-supplied ids on the coordinated path, add emitted-statement coverage for the saved round trips, and document the remaining managed-write costs and safe batching guidance.

Part of #533.


pdlug:

PR #540, opened 2026-08-22
perf(store): reuse negative endpoint probe

pdlug:
Reuse the initial negative root lookup in `getOrCreateByEndpoints` as the dispatcher result for the create path. The authoritative in-transaction match-key read remains unchanged, preserving convergence and stale/caching-transport correctness while reducing the common no-match path from three endpoint-match reads to two.

The statement-count regression test failed at three reads when the reuse was deliberately removed and passes at two with the optimization restored.

Part of #533.


pdlug:

PR #541, opened 2026-08-22
perf(store): skip generated-id existence probes

pdlug:
Generated node IDs now bypass same-kind duplicate/tombstone reads and cross-kind disjointness reads, while caller-supplied IDs retain the existing checks and resurrection behavior. Combined with the preceding identity fast path, a successful generated-ID node create now performs zero `getNode` or identity-fold probes before insertion.

The statement-count test asserts zero reads; deliberately restoring the probes made it fail with one read, and restoring the optimization returns it to zero.

Part of #533.


pdlug:

PR #542, opened 2026-08-22
perf(store): lease schema fences within transactions

pdlug:
Portable `Store.transaction()` callbacks on bundled backends now acquire the schema-version fence lazily on their first managed write and reuse it for later writes on that exact transaction target and expected version. Read-only transactions add no fence, while adapter/adopted and custom transaction targets retain conservative per-write fencing.

The scope is intentionally transaction-local: no Store, backend, pool-connection, or cross-transaction cache is introduced. Portable transaction targets expose neither native SQL nor nested `transaction()`, so no savepoint can predate the leased lock; the existing raw-savepoint regression remains on the per-write path. Two writes now emit one fence instead of two, and disabling reuse makes the load-bearing count test fail.

Part of #533.

### DIFF
--- a/.changeset/transaction-schema-fence-lease.md
+++ b/.changeset/transaction-schema-fence-lease.md
@@ -1,0 +1,5 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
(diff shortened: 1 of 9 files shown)


pdlug:

PR #543, opened 2026-08-22
perf(store): insert claim-free supplied node ids first

pdlug:
Successful single-node creates with caller-supplied IDs now use INSERT ON CONFLICT DO NOTHING RETURNING when the identity has no uniqueness or disjointness claims and the backend supports the fast path. Fresh IDs therefore avoid the preflight getNode round trip; occupied and tombstoned IDs are classified with an in-transaction diagnostic read.

Batches, constrained identities, and custom backends retain the existing read/partition/resurrection pipeline. Bundled SQLite and PostgreSQL backends implement the optional insert-first operation, including PostgreSQL transaction usability after an occupied-slot result.

The statement-count regression was proven load-bearing by disabling insert-first and observing the expected extra getNode statement; the mixed tombstone/new bulk regression was likewise reproduced by enabling the fast path for batches.

Part of #533.


pdlug:

PR #544, opened 2026-08-22
perf(store): fuse edge endpoint checks into insert

pdlug:
Successful generated-ID edge creates for unconstrained cardinality-many kinds now validate both live endpoints inside one INSERT SELECT RETURNING statement on bundled SQLite and PostgreSQL backends. The common path drops from two endpoint reads plus one insert to one database statement.

An empty result remains deliberately ambiguous and triggers the existing ordered diagnostic reads, preserving source-before-target typed errors for missing and tombstoned endpoints. Caller-supplied IDs, constrained edges, converging creates, and custom backends retain the portable path.

The statement-count regression was proven load-bearing by disabling the fused operation and observing both endpoint getNode probes return.

Part of #533.


pdlug:

PR #545, opened 2026-08-22
perf(store): guard constrained edge claims in one statement

pdlug:
Successful generated-ID constrained edge creates now replace the standalone cardinality probe with a guarded claim that checks both the claim holder and any pre-existing claimless incumbent. Combined with endpoint-validated insertion, the first-party SQLite and PostgreSQL path drops from five relevant database statements to two.

The guarded takeover checks the full declared cardinality axis, including the distinct one, unique, and oneActive liveness rules. Custom and legacy transaction targets retain the portable probe-then-claim protocol, and direct writers that bypass both the TypeGraph graph lock and claim relation remain outside the enforceable cardinality contract.

The statement-count regression was proven load-bearing by disabling guarded selection and observing the successful path return to five statements.

Part of #533.


pdlug:

PR #546, opened 2026-08-22
perf(store): fuse schema fences into inserts

pdlug:
Eligible first-party node and unconstrained edge creates now fold the active-schema fence into the INSERT itself. This removes one database round trip from the successful managed-write path while retaining PostgreSQL FOR SHARE locking and SQLite transaction-fence semantics.

Factory-owned root backends and transactions carry non-forgeable origin evidence through backend derivation. Custom backends, adopted transactions, and custom transaction wrappers keep the portable fence-then-insert path instead of silently opting into dialect-specific locking assumptions. A zero-row fused result uses a non-locking active-schema diagnostic, preserving the settled actual version without taking a second PostgreSQL lock.

The regression coverage observes actual driver SQL on SQLite and PostgreSQL, exercises custom-wrapper fallbacks for node and edge writes, and preserves duplicate, missing-endpoint, tombstoned-endpoint, stale-version, recorded-capture, and transaction-resource behavior. Removing the first-party eligibility marks makes the one-statement assertions fail, proving the statement-count property is load-bearing.

Part of #533.


pdlug:

PR #547, opened 2026-08-22
perf(store): autocommit atomic single-statement writes

pdlug:
Eligible schema-managed generated-ID node and unconstrained edge creates now execute as one autocommit SQL statement instead of BEGIN, fused INSERT, COMMIT. On high-RTT PostgreSQL links this removes two more sequential round trips from the successful path.

The selector is deliberately narrow: only exact bundled root backends may opt in, and node writes with identity, uniqueness/disjoint claims, searchable or embedding sidecars, history, or revision tracking remain transactional. Edges must be generated-ID cardinality-many creates with no convergence. Caller transactions, adopted handles, derived/custom backends, supplied IDs, batches, and all multi-statement plans retain the existing transaction path.

Driver-level statement assertions prove eligible SQLite and PostgreSQL/PGlite writes emit exactly one INSERT and no BEGIN or COMMIT. Inverting the generated-ID guard makes both the trace and selector tests fail, demonstrating that the fast-path boundary is load-bearing.

Part of #533.

### DIFF
--- a/.changeset/autocommit-single-statement-writes.md
+++ b/.changeset/autocommit-single-statement-writes.md
@@ -1,0 +1,5 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Avoid explicit transaction framing for bundled-backend schema-managed generated node and unconstrained edge creates when their complete write contract is one fused SQL statement.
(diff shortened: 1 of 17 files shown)


pdlug:

PR #548, opened 2026-08-22
perf(store): combine schema and graph write fences

pdlug:
Schema-managed PostgreSQL writes that also require the per-graph write fence now acquire the schema row and graph advisory lock in one ordered SQL statement. This removes one sequential round trip from constrained, history, and revision-tracked write plans while retaining the existing portable two-statement path for SQLite and custom backends.

The PostgreSQL statement uses dependent `MATERIALIZED` CTEs so the active schema row is locked `FOR SHARE` before the graph advisory lock can run. A zero-row result acquires no graph lock and uses the existing non-locking settled-version diagnostic, including the rollback-to-expected case where rejection still stands because no fence was acquired.

The combined acquisition seeds recorded capture's transaction-local lock memo, preventing history writes from paying a second reentrant advisory-lock round trip. Omitting that memo handoff makes the captured driver trace emit two graph-lock statements instead of one.

Part of #533.


pdlug:

PR #549, opened 2026-08-22
perf(store): fuse constrained edge claims and inserts

pdlug:
PostgreSQL and PGlite constrained-edge creates now combine live endpoint validation, cardinality claim acquisition, and the edge insert in one transaction-scoped statement on the common success path. This removes another sequential round trip after the schema/graph fence while SQLite and custom backends retain the portable guarded-claim plus insert flow.

Dependent materialized CTEs make live endpoints and an unoccupied declared axis prerequisites for touching the claim relation. Missing endpoints, claimless incumbents, mismatched low-level claim arguments, and duplicate edge identities therefore cannot leave a claim behind even when a caller catches the operation error and commits its surrounding transaction.

A foreign claim remains unresolved in the fused statement. TypeGraph reruns the existing ordered endpoint and cardinality diagnostics, then uses the guarded takeover on a fresh statement when the holder is stale, preserving PostgreSQL READ COMMITTED semantics for stale holders and lock-free writers. History capture touches the returned edge only when the fused insert succeeds.

Part of #533.


pdlug:

PR #550, opened 2026-08-22
perf(store): fuse node and fulltext writes

pdlug:
PostgreSQL and PGlite generated-ID node creates now derive their fulltext action before row work and apply the node insert plus tsvector projection from one `INSERT ... RETURNING` CTE. Searchable nodes with content upsert the projection, while all-empty searchable nodes run the delete arm from the same inserted-row source; a node conflict or projection failure leaves neither half applied.

The schema-fenced variant keeps the active schema predicate in that same statement. On the factory-owned root backend, an otherwise plain searchable create now qualifies for the existing autocommit single-statement plan, reducing the common path from `BEGIN`, schema-fenced node insert, fulltext write, and `COMMIT` to one database command. Transactional writes still preserve uniqueness claims, embeddings, recorded history, revision semantics, and contribution-initialization diagnostics around the fused row unit.

The capability is optional and supplied only by PostgreSQL fulltext strategies that can express the projection in one statement. SQLite FTS5, custom strategies, projected backends, caller-supplied IDs, and resurrection paths retain the existing portable write and sidecar flow.

Part of #533.


pdlug:

PR #551, opened 2026-08-22
perf(store): fuse generated node projections

pdlug:
## Summary

Replace the growing matrix of node-plus-sidecar backend methods with one typed, identity-free node projection plan. PostgreSQL and PGlite lower the generated node row, fulltext projection, and every present pgvector field into one materialized-CTE statement, including mixed searchable/vector schemas.

Batch durable contribution readiness for the whole projection plan into one cold-cache marker read, while the warm path performs only the atomic write. Root autocommit remains restricted to factory-owned backends and exact all-or-nothing strategy support; caller IDs, resurrection, history/revision guards, SQLite, and custom strategies retain the transactional fallback.

A warm generated searchable/vector create now uses one database command instead of BEGIN, the node insert, one fulltext write, one write per vector field, and COMMIT. A cold create adds one batched contribution-marker read.

Preserve typed schema-fence, missing-contribution, duplicate, and embedding-dimension diagnostics. The statement-count coverage was mutation-checked by disabling embedding projection fusion and observing the generated-vector fast-path case fail with two statements before restoration.

Part of #533.


pdlug:

PR #552, opened 2026-08-23
perf(store): fuse node constraint claims

pdlug:
## Summary

Extend the declarative node insert plan so one PostgreSQL statement can apply canonical pre-insert and post-insert uniqueness/disjointness claims together with the node row and any fulltext or vector projections. This removes one stateful round trip per claim while preserving the existing read probes for the next optimization batch.

The store computes claim placement and ordering once, then either lowers the complete plan through a transaction-scoped PostgreSQL backend or uses the existing portable standalone-claim fallback. An explicit atomic-claim capability prevents projected and third-party backends from silently accepting plan fields they do not implement.

Claim conflicts return structured holder metadata from the statement and preserve the existing UniquenessError and DisjointError contracts. Claim-bearing plans are refused on root autocommit handles and schema-fenced plans because their rollback boundaries cannot safely honor a post-DML refusal. The public claims field remains optional for source compatibility with projection-only plan callers.

The generated SQL dependency-links pre-claims, the node insert, post-claims, and projections. Runtime regressions cover successful mixed plans, fulltext composition, portable fallback, root and schema-fence refusal, and rollback after both pre-claim and post-claim conflicts. As a load-bearing mutation check, removing claims from the plan made the fused-statement regression fail on the missing uniqueness relation; restoring the plan returned it to green.

Part of #533.

### DIFF
--- a/.changeset/fuse-node-claim-write.md
+++ b/.changeset/fuse-node-claim-write.md
@@ -1,0 +1,5 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Fuse PostgreSQL node uniqueness and disjointness claims into the planned node insert, reducing constrained single-node writes by one round trip per claim while preserving portable backend fallbacks.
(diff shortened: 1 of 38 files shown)


pdlug:

PR #553, opened 2026-08-23
perf(store): compile managed creates into authoritative write plans

### LINKED ISSUE #530: getOrCreateByEndpoints match-key reads run outside the write transaction — stale-read transports break read-own-writes, and the failure blames a phantom competing writer

### LINKED ISSUE #533: Single managed write plans cost 5-6 round trips each — fence probe and duplicate checks dominate on high-RTT links

pdlug:
Replace the managed-create probe/write pipeline with authoritative plans that the bundled PostgreSQL backends lower into atomic statements. The store computes one claim and projection decision, threads it through every single-create/import fallback, and lets the backend return the database's verdict instead of re-deriving it from stale application reads.

## Architecture

Node creates now use `NodeCreatePlan` through `executeNodeCreatePlan`, covering schema fencing, uniqueness and disjointness claims, legacy-axis/live-row conflict checks, fulltext projections, and pgvector projections in one immutable unit. This is an intentional pre-1.0 source break from `NodeInsertPlan` and `insertNodeWithProjections`; custom backends retain the portable fallback, and existing persisted claim rows remain supported.

Edge creates reuse the same direction without prematurely translating arbitrary `matchOn` semantics into SQL: endpoint liveness, schema fences, and cardinality claims are folded into guarded inserts, while match-key reads remain transaction-scoped and authoritative. Root reads are dispatcher hints only, so caching transports such as Hyperdrive cannot turn a stale positive or negative into the operation's verdict.

TypeGraph-owned Store transactions lazily lease their schema fence. The first eligible planned write still embeds the fence in its own statement, preserving the one-write fast path; later node and edge writes reuse that exact transaction-local lease. Custom, adapter-native, and caller-adopted transaction targets retain conservative per-write fencing.

## Latency changes

| Managed path | Resulting database work |
| --- | --- |
| Generated claim-free node | Skips identity/existence probes and performs the schema-fenced insert directly |
| Generated searchable/vector node | Node plus every supported fulltext/vector projection in one statement |
| Generated unique/disjoint node | Claim verdicts, legacy/live conflict checks, node, and projections in one planned statement after any required graph lock |
| Generated edge | Endpoint checks are predicates of the insert rather than two preceding reads |
| Cardinality-constrained edge | Endpoint checks, guarded cardinality claim, and insert share one statement |
| Endpoint get-or-create miss | Reuses the negative root dispatcher read, confirms absence in the write transaction, then uses the fused endpoint insert |
| Multi-write Store transaction | One schema fence for the transaction, with no extra statement on its first eligible write |

`getOrCreateByEndpoints` now converges directly from rows observed by its transaction guard, confirms stale root positives before returning or choosing an update/resurrection id, and reports an unstable match key honestly instead of asserting a competing writer the database did not observe.

The round-trip regressions are load-bearing: removing negative-probe reuse changes the pinned endpoint-read count from two to three, removing leased-fence memoization changes the pinned fence count from one to two, and restoring external node claim probes breaks the exact planned-statement counts.

Closes #530
Closes #533


pdlug:
Superseded by #554, which targets main and contains the final node-and-edge planned-write architecture. The intermediate specialized edge hooks in this PR are removed there, so reviewing or merging this branch separately would create avoidable transitional API churn.

PR #572, opened 2026-08-28
fix: refuse dishonest Store transactions on transactionless backends

pdlug:
`Store.transaction()` and `Store.transactionWithReceipt()` now fail closed when a backend declares `capabilities.transactions: false`. Previously, both APIs executed their callbacks directly on the root backend, silently turning an atomic transaction contract into sequential, independently committed writes on transports such as Cloudflare D1 and Neon HTTP.

The refusal happens before callback invocation and uses `UnsupportedBackendCapabilityError`, so callers can make an explicit choice between a transaction-capable backend and deliberately non-atomic ordinary Store writes. The obsolete fallback implementation and its misleading transaction receipt behavior are removed.

Documentation now consistently describes transactionless Store behavior, updates serverless and event-log guidance, and includes migration guidance for callers that previously relied on the fallthrough.

### DIFF
--- a/.changeset/transaction-boundary-honesty.md
+++ b/.changeset/transaction-boundary-honesty.md
@@ -1,0 +1,5 @@
+---
+"@nicia-ai/typegraph": patch
+---
+
+Make `store.transaction()` and `store.transactionWithReceipt()` fail closed on backends without transaction support instead of invoking callbacks with non-atomic write semantics.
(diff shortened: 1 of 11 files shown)


pdlug:
Superseded by #574, which incorporates the transaction-boundary honesty fix into the consolidated atomic execution contract and convergence work.

PR #573, opened 2026-08-28
perf: converge durable edge batches as atomic mutation programs

pdlug:
Eligible durable `bulkGetOrCreateByEndpoints()` calls now execute as one schema-fenced atomic transport submission on bundled Neon HTTP, Cloudflare D1, and libSQL roots. The program lowers the whole unique batch into two shared multi-row statements: one endpoint-validated durable identity arbitration/write and one tombstone refusal guard. It returns ordered created/found results without the former outside probe, transaction framing, in-transaction reread, or per-item write loop.

This extends the existing `mutateEdges` capability as a discriminated mutation family instead of adding another backend seam. Eligibility is exact-root, fail-closed, based on unique durable identities, and derived from each transport's bind budget. Dynamic match fields, constrained cardinality, update and temporal options, caller transactions, custom or derived backends, and history/revision stores retain the complete portable path. That portable path now batches compatible existing-edge updates rather than issuing one write per row.

Native tombstone resurrection deliberately remains out of scope because TypeGraph cannot reproduce arbitrary Zod update transforms in SQL. A tombstoned winner rolls the native attempt back; transaction-capable backends use the schema-aware portable path, while transactionless roots refuse with the existing typed constraint error rather than weakening whole-call atomicity.

The implementation also restores input order independently of SQL return order, rejects partial or malformed program results, correlates endpoint refusal per input row, fails closed on stale schema fences and disabled durable-identity capabilities, and keeps every compiled statement within the backend's declared bind ceiling. Documentation now distinguishes the native live envelope from fallback and refusal states.


pdlug:
Superseded by #574, which includes this durable convergence implementation together with the execution capability contract, conformance suite, and transaction-boundary fix.

Issue #655, opened 2026-09-09
forked-working-copy Postgres test times out under load (CREATE DATABASE ... TEMPLATE)


pdlug:
## Observation

`tests/backends/postgres/forked-working-copy.test.ts` › "forks the base database by template, merges a fork write back to the base" times out at its 15 s budget under load, both on `main`'s CI (run 34188722929, PostgreSQL shard 1/2, commit f6d5387e) and on every full `pnpm test:postgres` run against a local server while other suites are active. Run alone it passes in a few seconds, every time.

The test issues `CREATE DATABASE ... TEMPLATE` for the fork, which PostgreSQL serializes against every other connection to the template database and which copies the whole template on disk; under a concurrent shard that is regularly slower than 15 s.

## Suggested fix

Either raise this suite's per-test budget the way the other in-process server suites do (`testTimeout` in the Vitest project config rather than the CLI, per the workflow contract), or provision the fork from a dedicated, minimal template database so the copy is small and uncontended. It should also be checked whether the template connection is left open by the harness between tests, since that alone blocks `CREATE DATABASE ... TEMPLATE`.

pdlug:
Fixed on the integration branch for #637 (no release carried the defect); closing now rather than waiting for the merge to main.

PR #677, opened 2026-09-11
fix(graph-merge): fork a PostgreSQL working copy without waiting on the template's checkpoint

pdlug:
`tests/backends/postgres/forked-working-copy.test.ts` timed out at its 15 s budget on every full `pnpm test:postgres` run and on CI's PostgreSQL shards, while passing in a few hundred milliseconds when run alone. The issue guessed at `CREATE DATABASE ... TEMPLATE` being serialized against the template's other sessions. It is not: measured against a PostgreSQL 18 lane server mid-run, the template copy costs 43-73 ms, and the 15 s is spent in `DROP DATABASE`.

`DROP DATABASE` forces an immediate, cluster-wide checkpoint and waits for it, because PostgreSQL has to know the checkpointer has forgotten the dropped database's sync requests before the files go away. Its duration is therefore set by how much every *other* suite in the lane has written since the last checkpoint, not by the size of the database being dropped. Dropping a 7 MB database freshly copied from a template took 13.7 s mid-run, of which 12.3 s was the checkpoint's fsync phase (`pg_stat_checkpointer.sync_time` rose by 12,253 ms and `num_requested` by exactly one); the next `DROP DATABASE` moments later, with the cluster already flushed, took 0.6 s, and the same drop on an idle server takes 0.1 s. That is why the test passes alone: alone, there is nothing to flush.

The suite needed two drops and both sat in the timed test body — one clearing residue before the copy, one inside the fork handle's `dispose`, reached through `close()`. Both now run in `beforeAll`/`afterAll` under an explicit 120 s host-DDL budget, leaving the body with the statement it is actually about. Clearing residue from an earlier run is harness setup rather than part of the fork, so `fork()` is now just the template copy; releasing the forked database is teardown, and `dispose` is still the thing that does it. Raising the per-test budget instead would only have moved the same unbounded wait behind a larger number, so the test budget is unchanged at the project default.

Relocating the drops must not cost the suite its proof that `dispose` runs, so the teardown reads `pg_database` and fails when the forked database survives `close()`, and the body asserts the copy exists in the first place. Both are guards that throw rather than `expect` calls, because `expect` outside a test block reports nothing useful from a hook and, for the name check, because the concrete forked name is what protects the `DROP` from ever targeting the suite's own live database.

No workflow change is needed: the PostgreSQL shards just invoke the lane script, and the cause is independent of worker count — the local reproduction runs one worker and CI's shard runs four.

`docs/TESTING.md` gains the measured fact under the per-suite-database notes, since any future suite that creates a database of its own faces the same trap.

Closes #655

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -252,6 +252,19 @@ Consequences worth knowing:
   server carries no state a run depends on — the per-suite databases are
   dropped and recreated — but leftovers from inspection do accumulate until
   the container is recreated.
+- **`DROP DATABASE` is a cluster-wide checkpoint, so it belongs in a hook, not
+  in a test body.** PostgreSQL forces an immediate checkpoint and waits for it
+  before removing a database's files, so the statement's duration is set by how
+  much every *other* suite in the lane has written since the last checkpoint —
+  not by the size of the database being dropped. Measured mid-run on the
+  PostgreSQL 18 lane server: 13.7 s to drop a 7 MB database, 12.3 s of it the
+  checkpoint's fsync phase, against 0.1 s for the same drop on an idle server.
+  A suite that creates a database of its own (only
+  `tests/backends/postgres/forked-working-copy.test.ts` does) therefore runs
+  both its drops in `beforeAll`/`afterAll` under an explicit hook timeout and
+  keeps the timed test body to the statement it is actually about. Raising the
+  test budget instead would only move the same unbounded wait behind a larger
+  number.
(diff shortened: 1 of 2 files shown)


pdlug:
Superseded: #668 on main fixes the same forked-working-copy timeout by draining the template before forking and is already merged into the integration branch, where the PostgreSQL lane is green on every head since. This branch predates that merge and only restructured the same test.

PR #720, opened 2026-09-16
chore: retire released API surface exceptions

pdlug:
Remove the released 0.64.0 API-surface exceptions so new feature branches record only their own compatibility decisions.


pdlug:
Closing as obsolete: the released-baseline exceptions this PR removed are already absent from main. Main now contains only eight active exceptions introduced after the published baseline, so resolving the conflict by preserving current main would leave no semantic diff. Removing those active entries before the next release would fail the API-surface compatibility gate.

